PR #6, opened 2019-12-06
Project Happy thoughts - Sofie Nyblad

nyblad:
https://nyblad-happy-thoughts.netlify.com/

Felt really low in the middle of the week and that I never would get this done, but somehow after mixing help from Damien and Jennies codealongs I managed to create the app and fix some bugs.
I booked a 1:1 on tuesday to clear some things out and hopefully understand everything a bit more. Still thinking it's hard with how and when to pass props and when to use useEffect and get the right syntax. And I should have gone with fewer components so it would't get even more complicated. Lesson learned until I feel more secure in React. And maybe wait until after wednesdays sessions before starting with the code, so I don't feel I have to start over if I can't get further with my own messy code.

- Get this error in console, what does that mean?

> Uncaught (in promise) undefined (index):1

- How to get the like button pink after more than 1 like?


code/src/components/LikeButton.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import { Heart } from './Heart'
+import './likebutton.css'
+
+export const LikeButton = (props) => {
+
+  const handleOnclick = () => {
```
ericaglimsholt:
```suggestion
  const handleOnClick = () => {
```

code/src/components/ListThoughts.js
```diff
@@ -0,0 +1,32 @@
...
+      {props.thoughts.map((thought) => (
+
+        <article className="thought" key={thought._id}>
+
+          <div className="message">{thought.message}</div>
+
+          <div className="details">
+            <div className="likes-wrapper">
+              {/* Passing the id and onThoughtLiked to Likebutton component */}
+              <LikeButton id={thought._id} onThoughtLiked={props.onThoughtLiked} hearts={thought.hearts} />
```
ericaglimsholt: Here you could add a `onClick()` that adds a class to make the heart to become pink!

PR #8, opened 2019-12-08
Happy Thoughts-Jenny Jacobsson

jennyjacobsson:
https://jenny-project-11-happythoughts.netlify.com/

**My reflections on how this week's project turned out:** I think it was a fun project this week. I finally feel like I'm starting to understand the thing about fetching API:s , using state and where props comes from and so on. It's not crystal clear, but a bit less confusing then a few weeks back. And I think it's so much easier to get started when you have a design to follow. Also I'm really thankful for these codealongs, they help out a lot!

Have a great day! ☀️

### DIFF
--- a/code/src/NewPost.js
+++ b/code/src/NewPost.js
@@ -4,18 +4,19 @@ import { ThoughtsList } from "ThoughtsList";
 export const NewPost = () => {
   const [newThought, setNewThought] = useState("");
   const [thoughts, setThoughts] = useState([]);
+  const [loading, setLoading] = useState(true);
 
   useEffect(() => {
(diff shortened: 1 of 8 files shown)


code/src/NewPost.js
```diff
@@ -0,0 +1,97 @@
...
+    <section className="post-wrapper">
+      <form>
+        {loading && (
+          <h2>
+            Don't worry, happy thoughts coming your way
+            <span role="img" aria-label="heart">
+              ❤️
+            </span>
+          </h2>
+        )}
```
ericaglimsholt: I really like this!

PR #16, opened 2019-12-08
Happy Thoughts -Petra Brinck 

petrabrinck:
Link to my deployed project: https://gallant-heisenberg-

Really liked the project but we started out dividing components (again) which for me was a big misstake since I wasn't sure about how to do it. Damien helped us (orange) over the phone but we still couldn't solve it. Thought I'd learn how to work with divided components along the way but that was not the case. I got it mixed up but backed a few steps and got the text and buttons in place but no functions working except the fetch. 
I had too many divs so after the codealong Wednesday I changed that but I think it would have taken me a long time to figure out how to make all functions work without the Friday session. 
I did however understand more after these two codealongs but this was tricky.

I put the font family in index.html but it doesn't apply so I don't understand what I did wrong.
I also added a questions regarding the dark border (half circle) around my like button on stack Overflow but hopefully someone knows the answer to why its there when I inspect and not the deployed :-)


ericaglimsholt:
Hi! Can you look at you Netlify and link to the correct url?

petrabrinck:
Hi,
This is the link to the deployed site https://gallant-heisenberg-b2ecb5.netlify.com

code/src/App.js
```diff
@@ -1,9 +1,42 @@
...
+      .then(res => res.json())
+      .then(json => setThoughts(json))
+  }, [postedMessage])
+
+  const onFormSubmit = message => {
+    setPostedMessage(message)
+  }
+
+  const onLiked = (thoughtId) => {
+    console.log("Logging in the App.js", thoughtId)
```
ericaglimsholt:
```suggestion
```

PR #18, opened 2019-12-08
Happy Thoughts of Aleksandra

AleksandraMuller:
Link to my deployed project:

https://happy-thoughts-of-aleksandra.netlify.com/

I thought that this project was quite hard and that is what I like about it the most. Thanks to that, I have learned a lot as I needed to google and read lots of articles and codes to get a good understanding of useEffect hook and fetching API's. I have even learned a bit more about props and how to use them within different components.

I was writing and trying out a lot of different codes until something worked out. I am proud of a lot of different little parts in my code. I would definitely need to structure my code a bit better and think through how to build components.

How I could structure my code better, what should belong into one component. I would also like to know if App.js should not have that many information and if I should only connect components to it.

Best,
Aleksandra


code/src/Components/ButtonMessage.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+export const ButtonMessage = props => {
```
ericaglimsholt: Maybe `Button` would be a better name?

AleksandraMuller:
Hi Erica! This makes sense! Thank you for your detailed reply :) When I will have some time, I will try to implement that 💃

PR #30, opened 2019-12-10
Happy-thoughts by Joacim Nilsen

JoacimNilsen:
Link to my deployed project:
https://serene-blackwell-83b86d.netlify.com/

I am happy with how close to the design it is. It felt easy at the beginning, but then as I wrote more code there is so much to keep track of. Next week i really will try to break it down in to smaller components.
I would like to get some pointers on how to make the like button to stay pink after click. And maybe how to implement a function that makes it only clickable once instead of having that counter of how many times I have liked something.


code/src/components/HappyThought.js
```diff
@@ -0,0 +1,43 @@
+import React, { useState } from 'react';
+import moment from 'moment'; // this is to format the date
+import './happyThought.css';
+
+// const url = 'https://technigo-thoughts.herokuapp.com/';
```
ericaglimsholt:
```suggestion
```

PR #31, opened 2019-12-11
PR - Tiago Ferreira

tavferreira:
Link to my deployed project: https://amazing-kalam-09369f.netlify.com

My reflections on how this week's project turned out: Not sure if I totally get what useEffect does and when should be used. I'll try to revisit the videos and look some more over the internet to understand it a bit better. Nevertheless I think the project looks good and I'm proud I could make the loader to work all by myself :)

Things I'd like to have clarified or explained in more detail: The codealong was awesome for me as I didn't have any time last week, but as I state useEffect is still a mystery for me.


code/src/components/HappyThought.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import moment from 'moment'
+import './happyThought.css'
+
+export const HappyThought = (props) => {
+    console.log('props', props)
```
ericaglimsholt:
```suggestion
```

PR #32, opened 2019-12-11
Project Happy Thoughts - Angelica B

Angelicabrodin:
Link to my deployed project:
https://myhappythoughts.netlify.com/

I thought it was really tricky to wrap my head around the actual methods, because I did not remember our work with get and post from the last project, and also the headers and what they actuallt does. So I really needed to take it step by step following the code-along, and also, I found it really hard do get started before we had the code along on last wednesday. 
When I actually started to see the result and that I even managed to post my thoughts, that was such a cool feeling :) Like that I actually made that happen just with my code, so all of these projects are so much fun just because you start to understand the backside of all the apps you use on a daily basis, so it was a really fun project to build. But again, hard to get started just with the help of the instruction in github. 

I did not manage to clear my text when submitting my thoughts in the form.. I tried different options but kinda dropped it when I decided that it was good enough without it. Otherwise I am very happy with my work and I kinda understand what I was actually doing :)

### DIFF
--- a/code/src/HappyForm.js
+++ b/code/src/HappyForm.js
@@ -7,7 +7,7 @@ export const HappyForm = ({ onFormSubmit }) => {
     const handleSubmit = (event) => {
         event.preventDefault()
         onFormSubmit(message)
-        setMessage('')
+
     }
 
     return (
--- a/code/src/HappyForm.js
+++ b/code/src/HappyForm.js
@@ -7,18 +7,24 @@ export const HappyForm = ({ onFormSubmit }) => {
     const handleSubmit = (event) => {
         event.preventDefault()
         onFormSubmit(message)
+        setMessage('')
     }
 
     return (
         <form>
-            <h3>Post a happy thought!</h3>
-            <p>{message}</p>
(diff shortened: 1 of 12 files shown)


code/src/HappyForm.js
```diff
@@ -0,0 +1,31 @@
...
+        event.preventDefault()
+        onFormSubmit(message)
+
+    }
+
+    return (
+        <form>
+            <h5>Post a happy thought!</h5>
+            <textarea
+                rows='3'
```
ericaglimsholt: When working on a project try to use `'` or `"` - do not combine them if you don't need to. In this case - you can use `"` as you are using all over the project
```suggestion
                rows="3"
```

code/src/HappyThought.js
```diff
@@ -0,0 +1,34 @@
+import React, { useState, useEffect } from 'react'
+import './happyThought.css'
+import moment from 'moment'
+
+export const HappyThought = (props) => {
+    console.log('props', props)
```
ericaglimsholt:
```suggestion
```

PR #33, opened 2019-12-11
project-happy-thoughts

jensnilsen:
Link to my deployed project:
https://compassionate-panini-ddb369.netlify.com/
Fun assignment. learnt a lot but would've been so stuck if it wasn't for the code-alongs

### DIFF
--- a/code/src/Components/HappyThought.js
+++ b/code/src/Components/HappyThought.js
@@ -1,14 +1,35 @@
-import React from "react";
-import "./happyThought.css";
-import moment from "moment"
+import React from "react"
+import moment from "moment" // this is to format the date
+import "./happyThought.css"
 
 export const HappyThought = props => {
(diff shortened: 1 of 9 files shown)


code/src/Components/HappyThought.js
```diff
@@ -0,0 +1,35 @@
...
+      body: "",
+      headers: { "Content-Type": "application/json" }
+    }).then(() => props.onLiked(_id))
+  }
+
+  return (
+    <article className='happy-thought'>
+      <h3>{message}</h3>
+      <div className="happyThoughtFooter">
+        <p className="buttonWrapper">
```
ericaglimsholt: Oh, here you have changed the way you write the html. Here you use `"` and in `HappyForm` you are using `'`. Please stick to one way of writing the syntax in the project

PR #36, opened 2019-12-23
Happy-thoughts-Ivett

Lilivet:
Link to my deployed project:
https://happy-thoughts-ivett.netlify.com/

It took me longer than I thought. Probably because I was very busy at work and with the family. When I took the time to look at the videos and do the reading, I could manage to make it work. 

There are two things that I could not fix: 
1) There is a square around the likes´ heart button. I tried to fixed it with CSS, but I couldn't. Any ideas?
2) I could not make the forms stayed inside the main container. I used max-with and margin:auto, to make it center, but then I could not make the content fit to the size of the main container.

### DIFF
--- a/code/package-lock.json
+++ b/code/package-lock.json
@@ -8586,6 +8586,11 @@
         "minimist": "0.0.8"
       }
     },
+    "moment": {
+      "version": "2.24.0",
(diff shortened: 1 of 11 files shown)


Lilivet:
Thanks Jenny!

I appreciate your feedback!

I will play with the code :-)

Regards,
Ivett

PR #39, opened 2020-03-23
Project Happy Thoughts -Hanna Rüter (review wanted)

Ruterryter:
Link to my deployed project:
https://epic-shaw-0a70d8.netlify.com/

It was a fun project even though i do not fully grasp how to make a good structure out of it and it took some time to make the pieces fit together. I looked at how some previous students did on this project and that made me a bit confused as well. 


Ruterryter:
Hi, I yes, i got that in some browsers yesterday, but in the localhost it works. I cant see any error messages so I'm a bit lost were it goes wrong. When I tried it now the first message showed, and the second one showed up with a bit of delay... I did a small change in the posting.js. Does it work now for you? Or could you give me a hint on were the problem is? Should I import more things in the app.js ?

Ruterryter:
Hi Erica, now it seems to work since I added the event.preventDefault(); in the posting.js. 
Does it work for you as well? /Hanna

Ruterryter:
Hi Erica! 
Now I have fixed the like button and got it to work when page refreshes.

PR #40, opened 2020-03-23
Review Wanted: week 11: Happy thoughts - Bea

bealun:

Started by mapping through the posted messages the have a feed with content. Continued with a creating a basic textarea to handle the input message that should be posted. Struggled quite a bit with the the structure and order of the functions. Got completely stuck on getting the message to actually show up and get updated without having to refresh the page. Solved it by creating a function and new hook call in App.js. 

Absolutley thrilled that I had time to deep dive in the project - so much fun and a lot of struggles! But I have managed to solve every obstacle I came across and feel like I actually understand why my code works. 

With the time I had left I added a loader that will show if the page takes time to load the feed and also a tooltip for the like-button. Also added a animation on the like-button that wil make the heart spin 360degrees when clicked! If I had more time I would love to know how to make it work for different browsers! 

I left the code the way I thought moz/webkit should be able to work but it doesn't but have also tried a bunch of different combinations. Would love to know to make it animated in other browsers - if there's a way!

Deployed site: https://happy-thoughts-bealun.netlify.com/


code/src/App.js
```diff
@@ -1,9 +1,48 @@
...
+
   return (
     <div>
-      Find me in src/app.js!
+      <UserInput formSubmit={handleFormSubmit} />
+
+      {loading ? <div className='loading'><ScaleLoader color='black' /></div> : 
+      thoughts.map(thought => (
+        <AlreadyPosted key={thought._id} thought={thought} onLiked={onLiked} />
+      ))}
```
ericaglimsholt: Nice!

code/src/components/Form.js
```diff
@@ -0,0 +1,50 @@
...
+        onChange={event => setMessage(event.target.value)}>
+      </textarea>
+      <div className='wordCount'>
+        <p>{message.length}/140</p>
+      </div>
+      <div className='sending'>
+
+        <button onClick={handleSubmit}
+          type='submit' tabIndex='0' aria-label='Submit button'
+          disabled={message.length <= 4 || message.length >= 140 ? true : false}>
```
ericaglimsholt: Really nice to see this!

PR #50, opened 2020-03-25
camilla happy thoughts

camillagaus:
Link to my deployed project: https://amazing-borg-5e3b05.netlify.com/

My reflections on how this week's project turned out: it was a bit hard to get the post method to work, espacially with the like button but it turned out fine at the end. 

Things I'd like to have clarified or explained in more detail: I would like to know how to add a loader to the page when posting a new post,


code/src/App.js
```diff
@@ -1,9 +1,37 @@
...
+
   return (
-    <div>
-      Find me in src/app.js!
+    <div className="thoughts">
+      <MessageInput />{
+      thoughts.map(thought => (
+         <PostedThoughts key={thought._id} onLiked={onLiked} thought={thought} />
+     )) 
+     }
```
ericaglimsholt: Would nice to have some more work on the indention here so it becomes more easy to read!

code/src/MessageInput.js
```diff
@@ -0,0 +1,54 @@
...
+        <textarea 
+          type="text"
+          minLength="6"
+          maxLength="140"
+          className="input-field" 
+          type="text"
+          id="myThought"
+          required
+          onChange={handleChange}
+          tabIndex='0'>
```
ericaglimsholt: Seems like you are switching between `'` and `"`. It would be nice to choose one and stick to it :)

code/src/MessageInput.js
```diff
@@ -0,0 +1,54 @@
...
+          id="myThought"
+          required
+          onChange={handleChange}
+          tabIndex='0'> 
+        </textarea>
+        <p className="char-count">{count}/140</p>
+      </label>
+      <input
+        type="submit"
+        value= '❤️    Add message    ❤️'
```
ericaglimsholt: The spacing here cause a problem in the way that the button here isn't responsive. Maybe checkout making the hearts in css with `:after` and `:before` so you can make it easier to do it responsive?

PR #51, opened 2020-03-25
Code review wanted -- Josephine

prosperpine:
Link to my deployed project: https://wonderful-swirles-55fc34.netlify.com/

My reflections on how this week's project turned out:I'm pretty happy with how it turned out, also happy that I got to do a stretch goal for once (the character counter).


code/src/MyThought.js
```diff
@@ -0,0 +1,50 @@
...
+      .then((newThought) => {
+        setThoughts((previousThoughts) => [newThought, ...previousThoughts])
+        setMyThought('') // will empty the textarea when the form is submitted
+      })
+  }
+
+  const countCharacters = () => {
+    const textEntered = document.getElementById('myThought').value;
+    const counter = (140 - (textEntered.length));
+    const countRemaining = document.getElementById('countdown')
```
ericaglimsholt: As you are ending the other lines with `;`, it would be nice if this follow the same structure
```suggestion
    const countRemaining = document.getElementById('countdown');
```

code/src/index.css
```diff
@@ -5,9 +5,113 @@ body {
...
+} 
+}
+/** MEDIA QUERY DESKTOP **/
+
+@media(min-width:1024px){
+.thoughts {
+    width: 40%;
+} 
+}
+
```
ericaglimsholt: The media queries would look nice with some indention work. Now it's easy to read because there isn't so much in each query but if you added more lines it can be hard to read

code/src/Thought.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import moment from 'moment'
+import { LikeButton } from './LikeButton'
+
+export const Thought = (props) => (
```
ericaglimsholt: In case you didn't know - you don't need `()` around when you are only passing one argument
```suggestion
export const Thought = props => (
```

PR #52, opened 2020-03-25
Happy Thoughts - Anna G

annaelsamaria:
Link to my deployed project:
https://hungry-lewin-3e50d8.netlify.com/

It looks fine. It was a hard project for me so I'm glad it turned out ok. 

How to style the heart.. Jennie showed us but that didn't work for me. It does not look good on my phone.


code/src/LikeHeart.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import { Emoji } from "./Emoji"
+import './LikeHeart.css'
+
+
+export const LikeHeart = (props) => {
```
ericaglimsholt: In case you didn't know - you don't need the `()` when you are only passing one argument
```suggestion
export const LikeHeart = props => {
```

code/src/PostThoughts.js
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <form onSubmit={handleSubmit} className="post-thoughts">
+      <h1>What's making you happy right now?</h1>
+      <textarea
+        value={message}
+        tabIndex='0'
+        aria-label='Input textarea'
+        className="form-text"
+        maxLength="140"
```
ericaglimsholt: Here you use both `'` and `"`. It would be nice if you choose one way :)

PR #53, opened 2020-03-25
Happy Thoughts – Kajsa Holmström

crushoftheyear:
https://happythoughts.netlify.com/

I split it up in components, using props and state to keep track of the thoughts and number of likes. I also added a counter to visualize number of characters in the textarea. The submit button gets disabled if the user types less than 5 or more than 140 characters.


code/src/components/LikeButton.js
```diff
@@ -0,0 +1,30 @@
+import React, { useState } from 'react'
+import { Emoji } from './Emoji.js'
+
+export const LikeButton = ({ hearts, id, apiUrl }) => {
+
+  const [likes, setLikes] = useState(hearts)
+
+  const addLike = (event) => {
```
ericaglimsholt: In case you didn't know - you don't need the `()` when only passing one argument
```suggestion
  const addLike = event => {
```

code/src/components/Textarea.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+export const Textarea = ({ message, setMessage, checkMessageLength }) => {
+  return (
+    <label>
+      What's making you happy right now?
```
ericaglimsholt: Maybe the label could be a prop aswell?

PR #55, opened 2020-03-25
Project Happy Thoughts Emma

emmapre:
Link to my deployed project:
https://happy-thoughts-emma.netlify.com/

I think it was really tricky to understand how to make everything work, especially the like-button. I'm quite happy with how it turned out after all and I like that I managed to do the validation on the form and the loading spinner. 

Why my background change of the like-button doesn't work, I'm quite sure it is something with the props but I can't figure it out.

### DIFF
--- a/code/src/app.css
+++ b/code/src/app.css
@@ -1,0 +1,7 @@
+.app-container {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  font-family: 'Roboto';
+  font-size: 18px;
+}
--- a/code/src/app.css
+++ b/code/src/app.css
@@ -1,7 +1,16 @@
 .app-container {
-  display: flex;
-  flex-direction: column;
-  align-items: center;
   font-family: 'Roboto';
   font-size: 18px;
+  
+}
+
+@media (min-width: 450px){
+  body {
+    display:flex;
+    justify-content: center;
+  }
+
+  .app-container {
+    max-width: 500px;
+  }
 }
--- a/code/src/components/Heart.js
+++ b/code/src/components/Heart.js
@@ -1,0 +1,29 @@
+import React from 'react'
+import './heart.css'
+
+export const Heart = (props) => {
+  const heartUrl = `https://technigo-thoughts.herokuapp.com/${props.id}/like`
+
+  const handleHeart = () => {
+
+    fetch(heartUrl,
+      {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json'
+        },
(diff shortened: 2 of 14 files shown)


code/src/app.css
```diff
@@ -0,0 +1,16 @@
+.app-container {
+  font-family: 'Roboto';
```
ericaglimsholt: Would be good the have a fallback font here!

code/src/components/Heart.js
```diff
@@ -0,0 +1,32 @@
...
+          'Content-Type': 'application/json'
+        },
+        body: ''
+      }).then(() => {
+        props.onMessageLiked(props.id)
+      })
+  }
+
+  return (
+    < button
```
ericaglimsholt:
```suggestion
    <button
```

code/src/components/Heart.js
```diff
@@ -0,0 +1,32 @@
...
+      className={props.hearts > 0 ? 'liked' : 'unliked'}
```
ericaglimsholt: The class is being set here. So you can add the background color to `.liked`!

code/src/components/NewThought.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  return (
+    <form
+      className='thought new-thought'
+      onSubmit={handleSubmit}>
+      <h2>What's making you happy right now?</h2>
+      <textarea
+        type='text'
+        maxLength="140"
```
ericaglimsholt: Here you mix `"` and `'`. Would look better if you use one way!

code/src/components/Thoughts.js
```diff
@@ -0,0 +1,64 @@
...
+        messages.map(thought => (
+          <article className='thought'>
+            <p
+              className="thought-text"
+              key={thought.id}>
+              {thought.message}
+            </p>
+            <div className='thought-details'>
+              <div className='heart-div'>
+                < Heart
```
ericaglimsholt:
```suggestion
                <Heart
```

code/src/components/Thoughts.js
```diff
@@ -0,0 +1,64 @@
...
+                  id={thought._id}
+                  message={thought.message}
+                  onMessageLiked={onMessageLiked}
+                  hearts={thought.hearts}
+                />
+                <p>x</p>
+                <p
+                  key={thought.id}>
+                  {thought.hearts}
+                </p>
```
ericaglimsholt: Can't this be in the same `<p>`?

PR #57, opened 2020-03-26
Project_11-Happy_Thoughts - Pål Hamel 

palhamel:
Link to my deployed project:
https://upbeat-wing-e5182a.netlify.com/

Fun project! Got a late start because of "life" but then a few really focused days that turned out well.

Did the brief's color scheme but then when "bananas". Have tried to limit my code comments, also had a talk with my mentor about that and got some good info how to handle this.


code/src/components/LikeButton.js
```diff
@@ -0,0 +1,26 @@
...
+       {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: ""
+    }).then(() => {
+      props.onThoughtLiked(props.id)
+    })
+  }
+  return (
+    // <button className="like-button" onClick={handleClick}><Heart /></button>
```
ericaglimsholt: Seems like you got a working version underneath!
```suggestion
```

code/src/components/MessageInput.js
```diff
@@ -0,0 +1,58 @@
...
+      <article className="submit-form-card">
+        <form onSubmit={handleSubmit}>
+        <h3>What's making you happy right now?</h3>
+          <textarea
+          minLength="5"
+          maxLength="140"
+          rows="4"
+          className="form-text"
+          onChange={(event) => setMessage(event.target.value)}
+          required
```
ericaglimsholt: If this would be indented one step it would help with the structure!

PR #59, opened 2020-03-26
project-happy-thoughts – Jennifer

Jenniferot:
Link to my deployed project:
https://boring-hamilton-513392.netlify.com/

My reflections on how this week's project turned out: It turned out pretty well I think. But I find it really tricky to wrap my head around useEffect. When I felt I understood a bit of react, props and useState it suddenly got very complicated with useEffect. 

Things I'd like to have clarified or explained in more detail: I have to look closer on the fetches, where to do them and how many thens and returns I can use accordning to one fetch. Have to look closer at useEffect too.


code/src/Emoji.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+export const Emoji = props => (
+  <span
+    className="emoji"
+    role="img"
+    aria-label={props.label ? props.label : ""}
+    aria-hidden={props.label ? "false" : "true"}
```
ericaglimsholt, line 8: Love this!

PR #61, opened 2020-03-26
Happy-Thoughts week11

VickBe:
Link to my deployed project: https://trusting-leakey-6f3d85.netlify.com/

I'm so happy it turned out :) There are many things I still don't get... But hopefully, if I keep on trying...
My if statement to change the color on the like-button depending on number of likes is not working and I can't find the reason why.
Aslo. I tried to make the submit button have a disabled state, thats also not working.


code/src/app.css
```diff
@@ -0,0 +1,4 @@
+body{
+  background-image: url(/code/assets/bg.png);
```
ericaglimsholt, line 2: Hm I don't get any background image? :/

PR #62, opened 2020-03-26
Happy-project-by Tuliany

Tuliany:
Link to my deployed project:
https://condescending-hopper-08aa97.netlify.com

I'm not so happy with it, because I was almost done with everything, then something went wrong... my brain. I named my different const in different components with similiar names and got so confused I ended up redoing everything and basically just followed Jennies repo. 

For me this week was a set-down for my ego, but at least I learned to be more clarified in the name-giving and also that I need to learn more about props.


code/src/HappyFeed.css
```diff
@@ -0,0 +1,54 @@
...
+  box-shadow: 4px 4px rgb(90, 88, 88);
+  width: 300px;
+  height: 130px;
+  margin: 16px;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+	padding: 20px;
+	overflow-wrap: break-word;
+
```
ericaglimsholt, line 13:
```suggestion
  padding: 20px;
  overflow-wrap: break-word;
```

code/src/HappyFeed.css
```diff
@@ -0,0 +1,54 @@
...
```
ericaglimsholt, line 1: There's some wierd indention going on in this file that would be good to sort out!

PR #64, opened 2020-03-27
project-happy-thoughts

louiseeriksson:
Link to my deployed project:
https://lucid-wright-1e1cc8.netlify.com/

im HAPPY with the result!

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from 'react'
 import { ThoughtsList } from './ThoughtsList'
 import { ThoughtInput } from './ThoughtInput'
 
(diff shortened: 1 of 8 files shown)


code/src/App.js
```diff
@@ -1,9 +1,69 @@
...
+    fetch(THOUGHTS_URL)
+      .then((res) => {
+        return res.json()
+      })
+      .then(data => {
+        setThoughts(data)
+      })
+  }, [])
+
+  const sendHeart = (id) => {
```
ericaglimsholt, line 39: In case you didn't know - you don't need `()` if you are passing only one argument
```suggestion
  const sendHeart = id => {
```

PR #66, opened 2020-03-27
Happy Thoughts week 11 Lisa Sturegård

LisaStur:
Link to my deployed project:
https://compassionate-hamilton-850c95.netlify.com/

Bit of a roller coaster... Started of unusually structured and planned, only to mess things up by trying to implement solutions from Monday- and Wednesday code-alongs that I did not fully master. Went back to basics and my original structure and got it back together.


code/src/components/LikeHearts.js
```diff
@@ -0,0 +1,24 @@
...
+      body: '',
+      headers: { 'Content-Type': 'application/json' }
+    }).then(() => onLiked(_id))
+  }
+
+  return (
+    <section className="like-hearts">
+      <button className={hearts > 0 ? 'liked' : 'not-liked'} 
+        onClick={handleClick}>
+        <p><span role='img' aria-label='Heart'>❤️</span></p>
```
ericaglimsholt, line 19: Do you need both `<p>` and `<span>` here?

code/src/components/MessageInput.js
```diff
@@ -0,0 +1,43 @@
...
+    })
+  };
+  
+  return (
+    <section className="input-section">
+      <form className="message-card">
+        <p>What's making you happy right now?</p>
+        <textarea className="textarea"
+          rows= '2'
+          type="text"
```
ericaglimsholt, line 30: Here you use both `'` and `"`. Would look nicer if you choose one way of writing!

code/src/components/MessageInput.js
```diff
@@ -0,0 +1,43 @@
...
+          onChange={event => setMessage(event.target.value)}>
+        </textarea>
+        <h4>{140 - message.length}</h4>
```
ericaglimsholt, line 33: Hm I really don't think this needs to be a title. I instead would like to have the line 27 as a title and this as a paragraph. What do you think about that?

code/src/components/MessageInput.js
```diff
@@ -0,0 +1,43 @@
...
+        <button className="btn"
+          type="submit"
+          onClick={handleClick} 
+          disabled={message.length < 5 || message.length > 140}>
+          <span role='img' aria-label='Heart'>❤️</span> Send Happy Thought <span role='img' aria-label='Heart'>❤️</span>
```
ericaglimsholt, line 38: Tip! You can actually add the hearts in plane css if you want to try that out (instead of spans). Then you can add the hearts to `:before` and `:after` in the css

code/src/components/LikeHearts.js
LisaStur: Hm, for some reason the hearts are not centered when i remove the <p>, but not making much sense, will have a proper look! Thanks!

code/src/components/MessageInput.js
LisaStur: Absolutely, big thanks!

code/src/components/MessageInput.js
LisaStur: Very good point! Will swap!

code/src/components/MessageInput.js
LisaStur: Super! Will try it out, very good tip! :)

PR #70, opened 2020-03-27
Olle Alm - Happy Thoughts

ollealm:
Link to my deployed project:
https://olle-happy-thoughts.netlify.com/

Struggle with understanding useEffect while starting to get a general understanding of React, there's a lot to take in :) Did some experimentation with the styling, but the general feel is the same as the guideline.


code/src/MessageInput.js
```diff
@@ -0,0 +1,47 @@
...
+
+      .then((res) => res.json())
+      .then((newMessage) => {
+        setMessages((previousMessages) => [newMessage, ...previousMessages])
+      });
+  };
+
+  return (
+    <div className="thought input">
+      <p className="input-text">What's making you <br />happy right now?</p>
```
ericaglimsholt, line 29: This would suit as a title!

PR #84, opened 2020-10-29
Project Happy Thoughts - Vanessa Sue Smith 💌

VanessaSue27:
Link to my deployed project:
https://happy-vane-thoughts.netlify.app/

This was a super fun project to build and I had a good laugh throughout the week looking at the thoughts the class was sharing.
It was a bit of a puzzle and a lot of back and forth thinking in order to understand what was going on.
Since each component has a special functionality to it, it was important to structure well where to keep the main data, and where we didn't need to do a new Get and where to do a Post instead.
Steps on how I approached this project are listed on the ReadMe file.
I pushed myself to do some of the stretch goals and managed to get some of those working 💪
I tried to get the amount of Likes given out to be stored on localStorage but didn't get that to work: localStorage saved the data as a string and would only save the data from the previous session, not all sessions in total. Had to let it go...😅 But at least it saves the Likes given per session 💕

All in all, a better experience than last week's using React, will probably need to build a similar project once more to let the concepts really sink in.
//Vanessa.


annahasselstrom (approved):
Oh Vanessa, it is really difficult to add value to your work - because you are a bit further on your coding journey compared with me. Anyway, I have put your code next to mine and as we already talked about last week, we have different approaches. 

You have your requests in your components and I have mine in App.js. You have all your css in one file - I've put each components's css in their own files. You use window.location.reload to refresh and so did I until Maks said that React does this for us. I wish I could explain in detail why we shouldn't use that method to reload, but perhaps this is something you can bring up with our coaches?

We approach useEffect differently - I create a function in App.js that returns the function fetchMessages(), which I call before the return of jsx in App.js. And in my fetch, located in App.js, I make both  POST requests. So I guess I handle both GET and POST with one useEffect (calling the function fetchMessages() twice (once for each POST). 
In my components I only set the state, handle the state and return jsx. The logic (useEffect, fetch) is in App.js.

I'm slowly getting a better grasp as to when to use which approach. And in Happy Thoughts maybe the "keep-logic-in-app.js" is to prefer because the children are dependent on data from one another but have no connection, which means that the parent - App.js - needs to provide this - passing data between components and server (the source of one truth?). Whereas in this week's project the children need not be connected since the data is being fetched from two different urls that do not effect each other. 

So Vanessa, you have done an amazing job - as always. You can create your own logic because you understand the fundamentals (I'm so envious...). But I haven't written any more than this text above in your code. I have nothing to add. Maybe you should ask someone with the same knowledge level as you to also look at your code?

code/src/App.js
```diff
@@ -1,9 +1,71 @@
-import React from 'react'
+import React, { useState, useEffect } from "react";
+import { ThoughtsCard } from "components/ThoughtsCard";
```
EmOlsz, line 2: There is a convention which says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, `style.css` should be separated as well and located below local dependencies. Ideally, it should look like this :

```jsx
import React from 'react';
import ReactDOM from 'react-dom';

import MyComponent1 from './components/MyComponent1';
import MyComponent2 from './components/MyComponent2';
import MyComponent3 from './components/MyComponent3';

import './style.css';
```

So, final order is :

1. Outer dependencies
2. Local dependencies
3. Style file(s)

code/src/App.js
```diff
@@ -1,9 +1,71 @@
...
+    //Fetches data from the API: an array including all Thoughts and
+    //uses the setThoughts setter function to assign that data to the
+    //thoughts variable
+    fetch(FETCH_URL)
+      .then((response) => response.json())
+      //json is now the variable containing thoughts array we got from the API
+      .then((json) => {
+        //Filter only Thoughts that have some text in it
+        const filteredThoughts = json.filter((thought) => {
+          return thought.message !== "";
```
EmOlsz, line 20: It's not 100% necessary, our API is not allowing to POST empty messages either way.
On the other hand, it's good to have couple of safe checks, so at the end this code can stay :)

code/src/components/Form.js
```diff
@@ -0,0 +1,56 @@
...
+        //Condition to change text color to red when there are only 10 characters left to reach max
+        style={{ color: newThought.length > 130 ? "red" : "black" }}>
+        {140 - newThought.length} characters left
+      </p>
+      <button
+        className="form-button"
+        type="submit"
+        /* Condition to enable the submit button only when the thought has right amount of characters */
+        disabled={
+          newThought.length < 6 || newThought.length > 140 ? true : false
```
EmOlsz, line 44: When we want to set attribute to a boolean value, we can omit the part about returning `false` or `true`.

This is overcomplicated piece of code :

```jsx
<button
  className="button"
  disabled={message.length < 6 || message.length > 140 ? true : false}
>
  Click me!
</button>
```

This is shorter, more accurate version of the same code :

```jsx
<button
  className="button"
  disabled={message.length < 6 || message.length > 140}
>
  Click me!
</button>
```

It does work, because logical operations such as && or || are always processed into boolean values at the end :

```jsx
const message = 'I am a string longer than 6, but shorter than 140'

message.length < 6 // This line of code is processed into false, because length
                   // of message is higher than 6

message.length > 140 // This line of code is also processed into false, because
                     // length of message is lower than 140

// This line of code
message.length < 6 || message.length > 140

// is basically
false || false

// therefore it's
false

// Final result of
disabled={message.length < 6 || message.length > 140}

// is
disabled={false || false}

// so
disabled={false}
```

code/src/components/ThoughtsCard.js
```diff
@@ -0,0 +1,47 @@
...
+  //by doing a POST request
+  //using that thought's id we make sure we add the heart to the right thought
+  const handleLikes = () => {
+    //Stretch goal: created a click counter so we know how many times the Heart button
+    //has been clicked and send that data in the callback function to App component
+    let clicks = 0;
+    clicks += 1;
+    fetch(`https://happy-thoughts-technigo.herokuapp.com/thoughts/${id}/like`, {
+      method: "POST",
+      body: "",
```
EmOlsz, line 17: While executing POST request for `likes` amount, we don't actually have to implement `body` property inside options object. API expects from us only unique ID inside url, that's the essence of this request.

Instead of writing :

```jsx
fetch(`https://happy-thoughts-technigo.herokuapp.com/thoughts/${id}/like`, {
    method: 'POST',
    body: {},
    headers: { 'Content-Type': 'application/json' }
});
```

We can just write :

```jsx
fetch(`https://happy-thoughts-technigo.herokuapp.com/thoughts/${id}/like`, {
    method: 'POST',
    headers: { 'Content-Type': 'application/json' }
});
```

code/src/components/ThoughtsCard.js
```diff
@@ -0,0 +1,47 @@
...
+            //Condition to determine heart button's back color depending on the amount of hears it has
+            style={{ backgroundColor: hearts > 0 ? "#ffadad" : "#f2f0f0" }}>
+            <span role="img" aria-label="Heart emoji">
+              &#128151;
+            </span>
+          </button>
+          x {hearts}
+        </p>
+        {/* Using the Moment.js package to determine when thought was posted */}
+        <p className="time-created">{moment(timeCreated).fromNow()}</p>
```
EmOlsz, line 43: Line break would be a little bit more descriptive :

```
<p className="time-created">
    {moment(timeCreated).fromNow()}
</p>
```

code/src/index.css
```diff
@@ -1,13 +1,108 @@
...
+  background-color: #ffadad;
+  border: none;
+  padding: 10px;
+  border-radius: 30px;
+  font-family: 'Trispace', sans-serif;
+  font-weight: bold;
+  cursor: pointer;
+}
+
+textarea {
```
EmOlsz, line 41: Good job with styling of `<textarea />`, I tried to break your styles by resizing it, but you styled it very well ;)

code/src/index.css
```diff
@@ -1,13 +1,108 @@
...
+  width: 500px;
+  box-shadow: 5px 5px;
+  padding: 20px;
+  background-color: #ffadad;
+  font-weight: bold;
+}
+
+/*Media Queries*/
+/*Mobile*/
+@media (max-width: 666px) {
```
EmOlsz, line 85: Not very common breakpoint, if you thought about mobile break, I would suggest `768px`

EmOlsz (commented):
Great job (well, being honest, as always) Vanessa!
I see you chose approach with site reload over POST-GET relation. As covered multiple times already, it's OK, it's just another way of creating our app's architecture.
Your functions are clean and do not have any major issues, only small things pointed out in my comments. Styles are consequent and well suited.
At the end, it's very well done project, take a look on my comments, they might provide you with extra set of theoretical knowledge :)

PR #89, opened 2020-10-30
Happy thoughts

LillanEriksson:
Link to my deployed project:
https://nervous-mayer-699099.netlify.app/

I'm really happy with how fast I could get the styling to match the given design! I would have loved to spend even more time on this project and reaching all red and black requirements, but since I'm going away for the weekend that's not possible.

I did not really understand the part about handling errors.


rebkova (approved):
Very nice work, Lillemor! The site Iooks and works as it should and I really like how you made the button for posting popping up when the length of the thought met the min character limit. One tip I can give you: you can use word-wrap: break-word; at an appropriate class in your CSS to avoid the text running out of the text container. Keep up the good work!

code/src/App.js
```diff
@@ -1,9 +1,52 @@
...
+	}, []);
+
+	const fetchMessages = () => {
+		fetch(THOUGHTS_URL)
+			.then((res) => {
+				return res.json();
+			})
+			.then((data) => {
+				data.sort((a, b) => a.createdAt > b.createdAt);
+				const filteredMessages = data.filter((item) => item.message);
```
EmOlsz, line 21: I think you wanted to use `.map()`, since current `.filter()` method will return all messages (so, basically there is no difference between `data` and `filteredMessages`)

code/src/Form.js
```diff
@@ -0,0 +1,50 @@
...
+
+	const handleSubmit = (event) => {
+		event.preventDefault();
+		onMessageChange(newMessage);
+	};
+
+	return (
+		<div className="message-container form-container">
+			<form onSubmit={handleSubmit}>
+				<label htmlFor="input-text">
```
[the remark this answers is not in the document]
EmOlsz: 👍

code/src/Form.js
```diff
@@ -0,0 +1,50 @@
...
+						rows="3"
+						value={newMessage}
+						onChange={(event) => setNewMessage(event.target.value)}
+					/>
+				</label>
+				<button
+					className="submit-button"
+					type="submit"
+					disabled={
+						newMessage.length < 6 || newMessage.length > 140 ? true : false
```
EmOlsz, line 28: Or you can simplify it : `disabled={newMessage.length < 6 || newMessage.length > 140}`

code/src/ThoughtsList.js
```diff
@@ -0,0 +1,40 @@
+import React, { useState } from 'react';
+
+import Moment from 'moment';
```
EmOlsz, line 3: `import moment from 'moment'` would be more reasonable, because `moment` is not a component, therefore we should import it with lowercase letter

PR #90, opened 2020-10-30
Happy thoughts - Johanna 

idsintehittapa:
Link to my deployed project:
https://silly-carson-4bd53e.netlify.app

The project is
🔵  Blue Level (Minimum Requirements)**

- following the design as closely as possible
- listing the most recent thoughts at the top and older thoughts at the bottom (sorted)
- showing the content of the message and how many likes they've received
- has a form to post new thoughts
- implementing the heart button to send likes on a thought

🔴  Red Level (Intermediary Goals)**

- Showing a count below the form input that updates as the user types and shows how many characters are remaining.

I have also added web accessibility for screen readers and responsiveness. 

I am ok happy with the project this week. 
For the next react project I'll structure things differently (in apps) and avoid using window reload.


code/src/Components/LikedMessage.css
```diff
@@ -0,0 +1,40 @@
...
+  /* -webkit-appearance: none; */
+  background: rgb(250, 168, 168);
+  border-radius: 50%;
+  border: none;
+  height: 35px;
+  width: 35px;
+  margin: 4px 10px 10px 12px;
+}
+
+.not-liked{
```
the-j-curl, line 17: Instead of repeating the code for .liked and .not-liked (since the only difference is the background colour), you could style just the button element and then styled .not-liked to the required bg colour.

code/src/Components/LikedMessage.js
```diff
@@ -0,0 +1,41 @@
...
+    }).then(() => {
+      setLikes(likes + 1)
+      window.location.reload()
+    }
+    )
+  }
+  return (
+    <div className='heart-time'>
+      <button
+        className={hearts > 0 ? 'liked' : 'not-liked'}
```
the-j-curl, line 23: Nice use of the ternary operators here. (Impressed face!)

code/src/Components/MessageList.css
```diff
@@ -0,0 +1,49 @@
...
+  margin-right: 35%;
+}
+
+@media screen and (min-width:768px) {
+  .message {
+      width: 70%;
+  }
+}
+
+@media screen and (min-width:1024px) {
```
the-j-curl, line 45: Another way to tackle this would be to specify a max-width, on the .message css rule

code/src/Components/MessageList.js
```diff
@@ -0,0 +1,45 @@
+import React, { useEffect, useState } from 'react'
+
+import moment from 'moment'
+
+import { LikedMessage } from 'Components/LikedMessage'
```
the-j-curl, line 5: No need to mention components folder as you are already in the same folder path. So this could be written as: './LikedMessage'

code/src/Components/MessageList.js
```diff
@@ -0,0 +1,45 @@
...
+  const MessagesUrl = 'https://happy-thoughts-technigo.herokuapp.com/thoughts'
+  const [messages, setMessages] = useState([])
+
+  useEffect(() => {
+    fetch(MessagesUrl)
+      .then((res) => {
+        return res.json()
+      })
+      .then((data) => {
+        console.log(data)
```
the-j-curl, line 18: Remember to remove console.log's before submitting project

code/src/Components/MessageList.js
```diff
@@ -0,0 +1,45 @@
...
+        // filter out empty messages 
+        const filteredMessages = data.filter((message) => message.message)
+        // save data to state
+        setMessages(filteredMessages)
+      })
+  }, [messages]) // this updates the content of the page without reloading it!
```
the-j-curl, line 24: I don't think your useEffect is working correctly as there is an infinite loop coming from your console log on line 18. I think removing "messages" and having an empty array would fix this.

code/src/Components/MessageList.js
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <section className='message-wrapper'>
+      {messages.map((message) => {
+        return (
+          <article className='message' key={message._id}>
+            <h2 tabIndex='0'>{message.message}</h2>
+              <div className='heart-time'>
+              < LikedMessage
```
the-j-curl, line 32: Spotted a whitespace between the < and LikedMessage

the-j-curl (approved):
Hi Johanna,
Nice work with this project. You have a smart looking app and I think overall very clean code.
However I have highlighted a few points for you to look at.

- Your useState in the MessageList component is generating an infinite loop in the console. 
This needs to be amended, I think removing the state "message" should fix it.

- I noticed that you haven't always left a space between CSS rule and curly bracket. Try to be consistent here in future.

- You use an image opening and closing tag, when this could be done by just using a  self closing tag.

- Instead of windows location reload - settimeout that renders every 10 seconds for example could of been a solution for this.

Keep up the great work, and let me know if you have any queries on any of the comments. It has been fun to read your code.

Jamie

code/src/Components/HappyForm.css
```diff
@@ -0,0 +1,73 @@
...
+  box-sizing: border-box;
+}
+
+.form-wrapper {
+  align-items: center;
+  display: flex;
+  flex-direction: column;
+}
+
+form{
```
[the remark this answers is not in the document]
EmOlsz: Thank you Jamie! :)

code/src/Components/HappyForm.css
```diff
@@ -0,0 +1,73 @@
...
+  width: 80%;
+  margin: 5vh 0;
+  background: rgb(230, 229, 229);
+  border-style: solid;
+  box-shadow: 5px 7px;
+  font-family: 'Inconsolata', monospace;
+}
+
+h1{
+  margin: 15px 15px 0 15px;
```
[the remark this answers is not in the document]
EmOlsz: Definitely

code/src/Components/HappyForm.js
```diff
@@ -0,0 +1,59 @@
...
+export const HappyForm = () => {
+	const url='https://happy-thoughts-technigo.herokuapp.com/thoughts'
+	const [message, setMessage] = useState('')
+
+	// a submit function witch POSTs the text input
+	const handleSubmit = event => {
+		event.preventDefault()
+
+		//send a POST request
+	fetch(url, {
```
EmOlsz, line 15: It's hard to understand your code because of wrong indentation, let's fix that

code/src/Components/LikedMessage.css
```diff
@@ -0,0 +1,40 @@
+.heart-image{
+  display: flex;
+  justify-content: space-around;
+  margin-bottom: 2.5px;
+}
+
+.liked{
+  /* -webkit-appearance: none; */
```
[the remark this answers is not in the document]
EmOlsz: 👍

code/src/Components/MessageList.js
EmOlsz: Exactly what I wanted to say, great catch Jamie. Needs to be fixed

code/src/Components/HappyForm.js
```diff
@@ -0,0 +1,59 @@
...
+          rows='4'
+          value={message}
+					onChange={event => setMessage(event.target.value)}
+					aria-label='Type a message no shorter than 5 nor longer than 140 characters'
+				/>
+			<div className='post-message-wrapper'>
+				<button
+					className='submit-button'
+					type='submit'
+					disabled={message.length <= 5 || message.length > 140 ? true : false}
```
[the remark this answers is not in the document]
idsintehittapa: Ah, thanks Jamie! Fixed 👍

code/src/Components/LikedMessage.css
idsintehittapa: I tried doing that, but it did not work as I intended - so this was the solution for now :)

code/src/Components/MessageList.js
idsintehittapa: Fixed

idsintehittapa:
I've done the requested changes

PR #96, opened 2020-10-30
project-happy-thoughts - Jamie Cook

the-j-curl:
Link to my deployed project:
https://think-happy-thoughts.netlify.app/

I feel that my project feels messy, and my knowledge is patchy in terms of where to place fetch's, posts, functions and useState variables. I broke my app when I tried to implement the like featured due to two different maps. This caused my cards to show 20 x 20!

If I had more time I would like to rearrange my API calls so that they are all in the app component. But I'm scared I will break the app and cause myself more work! 
I would like to know how I can move the API POST in ThoughtsList.js to App.js. And if the way I have completed the project is bad practice?


code/src/index.css
```diff
@@ -5,9 +5,157 @@ body {
...
+.main-container {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  margin-left: auto;
+  margin-right: auto;
+}
+
+.send-thought {
+  border: 1px solid black;
```
thereselinden, line 32: hex color (#000) might be a better way instead of "black"

code/src/index.css
```diff
@@ -5,9 +5,157 @@ body {
...
+  width: 90%;
+  height: 135px;
+  margin: 30px 0;
+  padding: 10px;
+  box-shadow: 5px 5px 1px black;
+  background-color: rgb(240, 236, 236);
+  position: relative;
```
thereselinden, line 39: This can be removed, together with position absolut on button and character, you would then have to wrap button and character in a div and display them as flex.

code/src/index.css
```diff
@@ -5,9 +5,157 @@ body {
...
+  font-size: 12px;
+  font-weight: 650;
+}
+
+.red-length {
+  color: red;
+}
+
+.black-length {
+  color: #000000;
```
thereselinden, line 88: This can be abbreviated to #000

code/src/index.css
```diff
@@ -5,9 +5,157 @@ body {
...
+}
+
+.thoughts-section {
+  width: 95%;
+  align-items: center;
+  margin-left: auto;
+  margin-right: auto;
```
thereselinden, line 95: This could be written as `margin: 0 auto;`

code/src/index.css
```diff
@@ -5,9 +5,157 @@ body {
...
+}
+
+.thought-card {
+  display: flex;
+  flex-direction: column;
+  border: 1px solid black;
+  height: 135px;
```
thereselinden, line 102: If you remove the height, it will adjust depending on how long the message written is.

code/src/ThoughtsList.js
```diff
@@ -0,0 +1,41 @@
...
+      {
+        method: "POST",
+        body: "",
+        headers: { "Content-Type": "application/json" },
+      }
+    ).then(() => onLiked(_id));
+  };
+
+  return (
+    <section className="thoughts-section">
```
thereselinden, line 17: You don't need to have the section here, article would be enough.

code/src/index.css
```diff
@@ -5,9 +5,157 @@ body {
...
+  margin-bottom: 30px;
+  padding: 10px;
+  box-shadow: 5px 5px 1px black;
+  word-break: break-word;
+  background-color: #ffffff;
```
thereselinden, line 107: #fff

code/src/index.css
```diff
@@ -5,9 +5,157 @@ body {
...
+}
+
+.thought-text {
+  font-family: Courier New, monospace;
+}
+
+.time-text,
+.like-text {
```
thereselinden, line 115: If these are only used/styled once their className could be the same. 😃

code/src/index.css
```diff
@@ -5,9 +5,157 @@ body {
...
+  .main-container {
+    width: 390px;
+  }
+
+  .send-thought {
+    width: 358px;
+  }
+
+  .thoughts-section {
+    width: 380px;
```
thereselinden, line 153: You could also have set a max-width on the first css rule, then you would not have needed to add media queries to style this 😃

thereselinden (approved):
Hello J-curl 🏄🏻‍♀️

Really nice work this week 💯 
I tried to explain how to move the 'like' fetch to the app component, if you have any questions just ping me and we can try to solve this together 🙌

I left some tips on how to shorten some of your css styles. I saw that you used different approaches when setting colors, maybe try to be consistent on what to use.

code/src/App.js
```diff
@@ -1,9 +1,55 @@
...
-}
+    <main className="main-container">
+      <ThoughtsForm onThoughtChange={reachThoughtInput} />
+      {thoughts.map(thought => (
+        <ThoughtsList
+          key={thought._id}
+          message={thought.message}
+          onLiked={onLiked}
+          hearts={thought.hearts}
+          _id={thought._id}
```
EmOlsz, line 49: We can omit underscore on the left side as we define new prop : 
```
id={thought._id}
```

PR #97, opened 2020-10-30
Happy Thoughts - Happy Mind - Gabriella Bolin

Gabbi-89:
Link to my deployed project: https://happyminds.netlify.app/

My reflections on how this week's project turned out: It was a really though project and hard to understand the connection between the different props.


code/src/styles/happythoughts.css
```diff
@@ -0,0 +1,38 @@
+.thoughts-box {
+  border: 1px solid #000;
+  box-shadow: 6px 6px #000;
+  width: 315px;
```
petrame, line 4: Notice that the fixed width of thoughts-box and post-box makes them a bit to big when surfing on small screens like galaxy fold.

petrame (approved):
Great job with the survey Gabriella! You nailed the styling with different colors on the like-button, a post-button which changes appearance depending on the number of characters written and also the counter that change to red when too much is said. 🤩
Over all clean code with good naming and indentation.

code/src/Components/ThoughtsList.js
```diff
@@ -0,0 +1,73 @@
+import React, { useEffect, useState } from 'react';
+import moment from 'moment';
+import { Likeathought } from 'Components/Likeathought';
+import 'styles/happythoughts.css';
```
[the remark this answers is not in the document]
Gabbi-89: What is dependencies? Can you elaborate what you mean I shall do? I really appreciate this kind of feedback! :)

code/src/App.js
```diff
@@ -1,9 +1,16 @@
-import React from 'react'
+import React from 'react';
+
+import { ThoughtsList } from 'Components/ThoughtsList';
```
EmOlsz, line 3: components folder should not be written in uppercase, neither do created like that. Only component files should be written in uppercase

code/src/Components/Postthoughts.js
```diff
@@ -0,0 +1,58 @@
...
+      }
+      // And then reload the window so we see the new thought that was added
+      ).then(() => {
+        window.location.reload();
+        });
+  };
+
+  return (
+    <form className="post-box" onSubmit={handleSubmit} > 
+      <label id={id}>What's making you happy right now?</label>
```
EmOlsz, line 28: I think you meant `htmlFor` over here instead of `id` :)

code/src/Components/ThoughtsList.js
```diff
@@ -0,0 +1,73 @@
...
+
+export const ThoughtsList = () => {
+  const THOUGHTS_URL = "https://happy-thoughts-technigo.herokuapp.com/thoughts";
+  const [thoughts, setThoughts] = useState([]);
+  // Use [] in the useState since it is an array that we get in the get-request
+  const [addThought] = useState("");
```
EmOlsz, line 10: And where is the state handler for update?

code/src/Components/ThoughtsList.js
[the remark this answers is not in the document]
petrame: The convention says that the outer (located in node_modules) / local (components etc) dependencies and style files should be divided by an empty line, first outer, then local and last style files.

code/src/Components/ThoughtsList.js
Gabbi-89: I have done a new push now. It works fine without state handler, or how shall I create one? :) @EmOlsz

EmOlsz: Yep, I see, but there is not logical reason to have this `addThought` property in state if you never update it :D
In state we should have only values that are going to somehow change after some exact actions of ours.
In this example however you initalise you `addThought` as empty string and never update it. Moreover, in line 35 of this file you added `addThought` into dependency array if `useEffect()`, but since your state property will never change its value, your `useEffect()` will not be executed

Gabbi-89: Now I removed it @EmOlsz So I hope I did right now.. :)

EmOlsz:
Project is not fixed, approved :)

PR #101, opened 2020-10-31
Happy Thoughts by Anna Hasselström

annahasselstrom:
Link to my deployed project:
https://inspiring-lalande-9954d9.netlify.app/

It has been fun and challenging as usual. Especially challenging this week was to understand how the different approaches presented to us, could or could not complement each other. A bit confusing. 

And to whomever is code-reviewing - sorry in advance! I've written lots of comments. They are written to myself for me to understand my code better but all the comments might make the code a little hard to read for a reviewer.

Local state and when we update only in browser in relation to 'the one source of truth'. 🤔


README.md
```diff
@@ -1,13 +1,42 @@
...
+- what a component lifecycle means
+- how to use the useEffect hook in React to perform actions when components mount, unmount, or when state changes.
+
+In this project, I have also learnt to: 
+
+- fetch message from the API using a GET request
+- create and send new messages using POST with a JSON body
+- send likes back - when user like a message - to the API-server using another endpoint and the method POST but without a body
+
+In sum, this week has taught me how to use APIs in React, and how to fire requests within 'useEffect'. I have also learnt how to put the results from the API responses into React state to output in the browser. This week I have both sent and received data from an API-server.
```
AxelHammarback, line 22: Such great learnings! And all in a week's time! Well done, Anna 👏

README.md
```diff
@@ -1,13 +1,42 @@
...
+
+# Tools and methods I have used
+
+- npm Node.js
```
AxelHammarback, line 26: Just out of curiosity, how did you use Node.js in this project?

code/src/App.js
```diff
@@ -1,9 +1,99 @@
...
+      headers: { 'Content-Type': 'application/json' }
+    }).then(() => {
+      //onLiked(messageId);
+      fetchMessages();
+    });
+  };
+
+  // The function onLiked below updates the browser locally but does not POST
+  // hearts to the server. Can be used to avoid time lag - to show likes if the response
+  // from the fetch takes some time. (true?)
```
AxelHammarback, line 61: I think you're on to something here – it's fairly common for applications to _assume_ the request will work out without problems, and update it locally while at the same time sending to the server and waiting for the response. If, for some reason, the response would come back with a "fail" or similar, apps usually post "Something went wrong when liking the post", or similar. Good thinking!

code/src/components/MessageForm.js
```diff
@@ -0,0 +1,40 @@
...
+        <h3>Post a happy thought!</h3>
+        <textarea
+          rows="3"
+          value={newMessage}
+          onChange={(event) => setNewMessage(event.target.value)} />
+        <div className="form-footer">
+          <button
+            type="submit"
+            onClick={handleSubmit}
+            disabled={!!(newMessage.length < 5 || newMessage.length > 140)}>
```
AxelHammarback, line 32: Really smartly done with the validation – well done!

code/src/components/MessageForm.js
```diff
@@ -0,0 +1,40 @@
...
+            <span className="heart-emoji" role="img" aria-label="heart">❤️</span> Send Happy Thought<span className="heartr-emoji" role="img" aria-label="heart">❤️</span>
```
AxelHammarback, line 33: Is "heartr-emoji" a typo? I can't find it anywhere else...

code/src/components/messageForm.css
```diff
@@ -0,0 +1,141 @@
...
+  transition: 0.2s all; 
+  box-shadow: 7px 6px 28px 1px rgba(0, 0, 0, 0.24);
+}
+
+.happy-form button:hover {
+  cursor: pointer;
+}
+
+.happy-form button:focus {
+  outline:0 !important;
```
AxelHammarback, line 49: While the designer in me wants to do the same thing here (outline: 0), it's generally considered bad for accessibility purposes. 

I know accessibility isn't a requirement in this project, and if it's a conscious choice because of it, then ignore this comment. 

But in case you weren't aware of this, I just thought I'd let you know. See more here: http://www.outlinenone.com/

AxelHammarback (commented):
Anna, really well done with everything! I'm not only impressed by the clarity of your code, but also the way you document your learnings, and how well-commented it is. Really, really well done – such impressive progress 👏👏👏

code/src/components/MessageForm.js
EmOlsz: Indeed :) Btw, we don't need `!!` over here, whole logical operation `newMessage.length < 5 || newMessage.length > 140` will already return boolean value for us, two `!` marks will just do 360 degrees spin

PR #103, opened 2020-10-31
Cuz I'm Happy

karanmann:
# Happy Thoughts

We here have made a Twitter Clone using REACT.js where we can fetch messages posted on herokupp and also post them and add likes to the posts. Also we check how long back the post was made using MOMENT.js and added some validations depending on the length of the message.

## The problem

Initially I was planning on breaking down the project into smaller components but then decided against it since they were not getting reused. I did however use a different component to make a like button which was quite a challenge as I was not able to post but after an almost frustrating half day I finally had a light- bulb moment and was able to get it working.

## View it live

https://cuz-i-am-happy.netlify.app/


code/src/index.css
```diff
@@ -1,13 +1,65 @@
...
+  width: 20px;
+}
+
+.message-text {
+  word-wrap: break-word;
+}
+
+.message-heart {
+  display: flex;
+  flex-direction: column;
```
jespa374, line 28: I can't find this either. Is it used somewhere? Otherwise you should remove this

code/src/index.css
```diff
@@ -1,13 +1,65 @@
...
+}
+
+textarea {
+  font-size: 17px;
+  font-family: "Trispace", sans-serif;
```
jespa374, line 33: Good that you are using fallback font :)

jespa374 (approved):
Great job Karan! You have achieved the blue level requirements and some red level requirements as well. As commented ,there are some places you have issues with indentation which is maybe due to the problems in your editor. Also, you could check your counter validations as mentioned in the comments. You can take a look at the comments and do the changes proposed. Overall job well done ! :)

karanmann:
Thank you @jespa374 Jessica,

The indentation was because of the previous problem I had and it is now fixed. Also that you for remindind me about how the priorities of the imports work and I completly did miss the max length. Thank you again. Now the issues have been debugged :)

code/src/components/MessageList.js
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <div>
+      {messages.map((getMessages) => {
+        return (
+          <div className="message-card" key={getMessages._id}>
+            <p className="message-text">{getMessages.message}</p>
+            <div className="likes-time-container">
+            <LikeButton
+                key={getMessages._id}
```
EmOlsz, line 42: No need for `key` over here, you only need to add it to root element of this loop, so `.message-card` element :)

PR #105, opened 2020-10-31
HappyHappy

moablomkvist:
https://happymessageappmoa.netlify.app/

This site is connected to a server and we are fetching the api for GET and POST to develop this chat function of Happy Thoughts. 

## The problem

I started by collecting all the fetching in the app component, but I had to start again with a new approach because I lost myself when I had to solve the like button. Too bad, because I had something good going on.


code/src/components/Like.js
```diff
@@ -0,0 +1,31 @@
...
+
+import './like.css'
+
+export const Like = ({ onLiked, heart, heartId }) => {
+  const handleClick = () => {
+    fetch(
+      `https://happy-thoughts-technigo.herokuapp.com/thoughts/${heartId}/like`,
+      {
+        method: "POST",
+        body: "",
```
EmOlsz, line 11: No need for a body in this request, all essential informations are being passed in url in line 8

code/src/components/ThoughtsInput.js
```diff
@@ -0,0 +1,65 @@
...
+    fetch(URL, {
+      method: "POST",
+      body: JSON.stringify({ message }),
+      headers: { "Content-Type": "application/json" },
+    })
+      .then(() => {
+        setMessage("");
+        onFormSubmit(message);
+      })
+      .catch((err) => console.log("error:", err));
```
EmOlsz, line 20: or `.catch((err) => console.error("error:", err));`

PR #110, opened 2020-11-01
Give me some love - Andrea 

Andrea-Osmar:
Link to my deployed project: https://give-me-some-love.netlify.app/

My reflections on how this week's project turned out: I thought this weeks project was very hard and quite messy. I have a hard time understanding what I have created :/ 

Things I'd like to have clarified or explained in more detail: in my iphone the text "send a happy thought" is not in the centre. however it looks good in mobile vbiew in chrome and on other phones. How can i fix this?


Hanna-Skeppe (approved):
Andrea I think you did a great job with this project! 
I saw that you said in the README that you thought this project was hard and I agree that it was a lot to take in this week. 
I also saw you had trouble solving how to get the text on the "send happy thought" centered. I don't really have a solution to this, but it might have something to do with the hearts being bigger than the text? Maybe experiment some with padding/margin would do the trick?
I think your code was really clear and well-structured. And you have consistently good indentation and use of semicolons. That is a huge plus! I also think you had a good use of semantic tags throughout the code. Well done!
I added some comments that I hope are helpful. And I have a few suggestions on how you could improve the project:
-Accessibility: I tried to tab through with Chromevox and it didn't read the heading on the input form or the messages. So maybe add tabindex there for it to work.
-Styling: overall I think your styling works well on different devices. A suggestion is to maybe add some more media queries to make it even more responsive. But that is of course optional.
I enjoyed reading your code and I think you did a very good job. I see that you fulfill blue level and also some of the red level. 🥇

code/src/components/MessageList.js
```diff
@@ -0,0 +1,39 @@
+import React, { useState } from "react";
+import moment from "moment";
+
+export const MessageList = ({ messageList, onLikeChange }) => {
+  const [likes, setLikes] = useState(0);
+
+  const handleLike = (id) => {
+    setLikes(likes + 1);
+    onLikeChange(id);
+    console.log(likes);
```
EmOlsz, line 10: We should delete all console.logs before final push

code/src/components/MessageList.js
```diff
@@ -0,0 +1,39 @@
...
+  };
+
+  return (
+    <div>
+      {messageList.map((messages) => (
```
EmOlsz, line 15: it's a single message we take as an argument, this would be more appropriate :

```
{messageList.map((message) => (
```

PR #116, opened 2020-11-01
Week 11: Happy thoughts

monkeybusiness101:
Link to my deployed project:

https://sleepy-carson-4649f3.netlify.app

Fun but challenging project. I wished I knew that there were more videos to watch (Which Poya told us on Friday..)


code/src/Like.js
```diff
@@ -0,0 +1,18 @@
...
+        //POST like on specific comment
+        fetch(`https://happy-thoughts-technigo.herokuapp.com/thoughts/${_id}/like`, {
+          method: "POST",
+          headers: { "Content-Type":"application/json" },
+          body:"",
+        })
+    }
+
+  return  <button onClick={() => {handleLike(_id)}} className="like-button">
+            <span role="img" aria-label="Heart emoji"> ❤️ </span>
```
IngelaL, line 15: Here you could add and if statement that if the button gets 1 or more clicks, it should change background color. Like this: style={{ background: hearts > 0 ? '#feacac': 'f2f0f0'}} (Then you also have to add the hearts prop in the component). On your web page the background color of the heart only changes temporarily (when clicking on it) But then it changes back to the original background color.

code/src/index.css
```diff
@@ -1,13 +1,89 @@
...
+
+.like-button {
+  height: 35px;
+  width: 35px;
+  border-radius: 50%;
+  margin-right: 5px;
+}
+
+.like-button:focus {
+  background-color: #ffadad;
```
IngelaL, line 88: This only make the color of the like button change color when clicking on it but when you click utside of it it goes back to the background color it had before. Once the happy thought get a like, the color should change permanantly. See comment in LikeComponent line 15.

code/src/index.css
```diff
@@ -1,13 +1,89 @@
...
+.message {
+  width: 90%;
+  height: 150px;
+  border:solid 2px #000;
+  background-color: #c9d5e1c1;
+  box-shadow: 5px 5px;
+  margin-top: 50px;
+  font-size: 12px;
+  padding: 5px;
+}
```
IngelaL, line 60: Here you could add a 'max-width' of maybe 400px so the boxes don't get so stretched out on a big screen. Same with the form.

code/src/Fetch.js
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from 'react'
+import { useEffect } from 'react'
+import moment from 'moment'
+
+import { Like } from 'Like'
+
```
IngelaL, line 6: Great that you separate the react import from the components import.

code/src/Fetch.js
```diff
@@ -0,0 +1,50 @@
...
+        fetch('https://happy-thoughts-technigo.herokuapp.com/thoughts')
+        .then((response) => {
+            return response.json();
+        })
+        .then((text) => {
+            setMessages(text)
+            console.log('Request successful', text)
+        })
+        .catch((error) => {
+            console.error('Request failed', error)
```
IngelaL, line 22: 👍

IngelaL (changes requested):
Hi Max! You have a great structure of your components and your code looks great in general. What you need to do is to fix two things with the heart like-button: 1. to make the background stay red when clicked. 2. To make the "heart-count" work. When adding a like it does'nt update the number (x 1) on the page. (However, it works "behind the scenes"). I also have some suggestions on making your code more readable in some places.

monkeybusiness101:
@IngelaL Thank you so much for the feedback, that was awesome!
- I have removed unnecessary blank spaces and fixed some indentation
- I fixed so that the like button keeps it color if there are more than one like (thank you for that)
- I added so that like count is more dynamic

Let me know if you want me to have a look at anything else.

Great PR!

monkeybusiness101:
I made it look nicer on desktop as well (thank you for the input on that)

IngelaL:
Looks great now Max!

IngelaL (approved):

code/src/App.js
```diff
@@ -1,9 +1,13 @@
...
 
+import {Fetch} from 'Fetch'
+import {Create} from 'Create'
+
 export const App = () => {
   return (
     <div>
-      Find me in src/app.js!
+      <Create />
+      <Fetch />
```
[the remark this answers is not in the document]
EmOlsz: Agree!

code/src/Create.js
```diff
@@ -0,0 +1,36 @@
+import React, { useState } from 'react'
+
+export const Create = () => {
+    
+    const [text, addMessage] = useState("")
+
+    const handleSubmit = event => {
+        
+        event.preventDefault()
+        console.log(text)
```
EmOlsz, line 10: Let's delete all of those console.logs

code/src/Create.js
```diff
@@ -0,0 +1,36 @@
+import React, { useState } from 'react'
+
+export const Create = () => {
+    
+    const [text, addMessage] = useState("")
```
EmOlsz, line 5: Names of state and state handlers should have same base so it's easy to detect their pairs. In your case, correct naming would be :

```
const [text, setText] = useState("");
```

or 

```
const [message, setMessage] = useState("");
```

monkeybusiness101: Fixed

code/src/Create.js
monkeybusiness101: Fixed

code/src/App.js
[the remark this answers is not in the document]
monkeybusiness101: Fixed

monkeybusiness101:
@EmOlsz Hi Maks! I have removed console.logs, changed the components to reflect their purpose better, and changed the usestate to correlate.

/Max

PR #118, opened 2020-11-01
Happy thoughts project - Emma

emma-tech:
https://happy-thoughts-project.netlify.app/

I'm very happy with this weeks project! It was a lot of new and complicated things to learn, but I'm very happy with the result :) 

I had a lot of issues with deploying the site to Netlify, in the end I had to move the Styles-folder into the components folder (and of course update the paths accordingly).. but I've seen other examples where the Styles folder don't need to be in the components folder, in the previous project for example I just had the styles-folder in the src-folder. Why didn't that work this time?


code/src/App.js
```diff
@@ -1,9 +1,70 @@
...
+import React, { useState, useEffect } from 'react';
+
+import { ThoughtsList } from 'components/ThoughtsList';
+import { ThoughtForm } from 'components/ThoughtForm';
+import { Header } from 'components/Header';
+import { Footer } from 'components/Footer';
+import { ThoughtsUrl } from 'Urls';
 
 export const App = () => {
+
```
AnnaHellqvist, line 10: Great comments through the app.js file! really makes it easier to understand :)

code/src/components/ThoughtsList.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import moment from 'moment';
+
+import './Styles/thoughtslist.css';
+
+export const ThoughtsList = ({thought, onLike}) => {
```
AnnaHellqvist, line 6: Maybe reconsider naming of the component, since this is in fact a Thought and not the list? :)

code/src/App.js
```diff
@@ -1,9 +1,70 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react';
+
+import { ThoughtsList } from 'components/ThoughtsList';
+import { ThoughtForm } from 'components/ThoughtForm';
+import { Header } from 'components/Header';
+import { Footer } from 'components/Footer';
+import { ThoughtsUrl } from 'Urls';
```
AnnaHellqvist, line 8: add ./ in the beginning of file paths

code/src/App.js
```diff
@@ -1,9 +1,70 @@
-import React from 'react'
```
AnnaHellqvist, line 1: add ;

README.md
```diff
@@ -1,13 +1,22 @@
-# Happy Thoughts
+# Happy Thoughts 💌
```
AnnaHellqvist, line 1: Great readme, you have really explained your app and how you approach your problem :)

code/src/App.js
```diff
@@ -1,9 +1,70 @@
...
+import { Footer } from 'components/Footer';
+import { ThoughtsUrl } from 'Urls';
 
 export const App = () => {
+  
+  //State hooks for thoughts, starting out with an empty array
+  const [thoughts, setThoughts] = useState([]);
+  
+  useEffect(() => { // Adding use effect hook to control the fetch of happy thoughts from API
+    fetchThoughts();
```
AnnaHellqvist, line 15: I like this, looks more clean with the function below  👍

code/src/components/Header.js
```diff
@@ -0,0 +1,16 @@
...
+import './Styles/header.css';
+
+
+export const Header = () => {
+
+    return (
+        <h1 className='header'>
+                Share your happy thoughts
+            <div>
+                <span className='pulse-heart' role='img' aria-label='heart'> ❤️ </span>
```
AnnaHellqvist, line 12: This is really cute!! ❤️

code/src/components/ThoughtForm.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    // Form for posting happy thoughts
+    return (
+        <form onSubmit={handleSubmit} className='thoughts-form'>
+            <h2 className='thoughts-form__header'> 
+                What's making you happy right now?
+            </h2>
+                <input
+                    type='text'
```
AnnaHellqvist, line 24: when I input a long message in your form it only uses the first row. So when I keep typing I cannot see my whole message. Maybe try a <textarea> instead of < input type=text > or if the css property word-break: break-word works on this as well :)

code/src/components/Styles/thoughtform.css
```diff
@@ -0,0 +1,47 @@
...
+    background: #eeeeee;
+    border: 1px solid;
+    box-shadow: 5px 10px;
+    width: 100%;
+}
+
+.thoughts-form__input {
+    width: 100%;
+    padding: 5px 0px 80px 5px;
+
```
AnnaHellqvist, line 15: u could remove the line of whitespace and when there is a 0 value you can omit the px :)

AnnaHellqvist (approved):
Great job on this project Emma! Your code is clean, easy to read and you are really good at adding comments that provides extra understanding. You managed to name the CSS classes in a good way, it's easy to know what you mean. Is it BEM methodology? I had just a few comments you can take a look at. Be proud of your self and keep up the good work 😄 🥇

code/src/Urls.js
```diff
@@ -0,0 +1 @@
+export const ThoughtsUrl = 'https://happy-thoughts-technigo.herokuapp.com/thoughts';
```
EmOlsz, line 1: Since url is not a component, we should rather name it in lowercase :) This would be correct way :

```
export const thoughtsUrl = 'https://happy-thoughts-technigo.herokuapp.com/thoughts'; 
```

PR #122, opened 2020-11-01
Project Happy Thoughts - Hanna Skeppe

Hanna-Skeppe:
Link to my deployed project:
https://hannas-happy-thoughts-app.netlify.app/

It turned out ok. I did the blue and some of the red requirements. The hardest part was to understand the code I've written, how it actually works. Had a bit of a hard time to understand use-effect and use-state and what props to pass to where and where I should put different things in my code.
I tried to put a background image, but it didn't work. I didn't know where to place the background picture. It didn't work to place it on the body.

I would have liked to have the second red-level goal explained more: "When POSTing a new thought, if the message was empty, too long, or too short, you get an error message back from the API. Use this to set some sort of `error` state to show a friendly message to your user. (Hint: Use the network tab of the developer tools in your browser)".


code/public/index.html
```diff
@@ -4,31 +4,13 @@
...
-      Notice the use of %PUBLIC_URL% in the tags above.
-      It will be replaced with the URL of the `public` folder during the build.
-      Only files inside the `public` folder can be referenced from the HTML.
-
-      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
-      work correctly both with client-side routing and a non-root public URL.
-      Learn how to configure a non-root public URL by running `npm run build`.
-    -->
+    <link href="https://fonts.googleapis.com/css2?family=Inconsolata:wght@300;400;500&family=Roboto:wght@100;300;400;500&display=swap" rel="stylesheet">
     <title>Technigo React App</title>
```
trangtara, line 8: It is recommended to change the title into something more relevant to the Happy thought.

Hanna-Skeppe: Added title 👍

code/src/App.js
```diff
@@ -1,9 +1,66 @@
...
+  };
+
+  //fetch POST HappyLike:
+  const postThoughtLike = id => {
+    fetch(`https://happy-thoughts-technigo.herokuapp.com/thoughts/${id}/like`, {
+      method: 'POST',
+      body: '',
+      headers: { 'Content-Type': 'application/json' },
+    })
+      .then(() => fetchMessages()) //What is the difference between writing like this and just: .then(fetchMessages())?
```
trangtara, line 57: To access the error, I suggest we get the res.json. In the json, there will be json.errors with the information for us to handle. If there is any errors in the json, we will need to use the "throw" statement to provide the error message. Then, outside the fetch function, we will put the catch statement. One of the code options can be like this:
.then((res) => res.json())
.then((json) => {
   if (json.errors) {
      throw "Oops! Something went wrong"
    }

Outside the fetch function, we will have a catch statement to print out the alert message. I can be like this:
.catch ((error) => window.alert(error)).
Above is one simple solution that I came across with. If we want to print out different messages for different kinds of errors, we need to use State hook.
But I can suggest that you fetch res.json, json.errors to discover how the errors look like first. 
Long message I have typed here and maybe it is better to see the solution on real code. If you want, you can check my repo. I am sure there are different ways to solve this too.
https://github.com/trangtara/project-happy-thoughts/blob/master/code/src/App.js

trangtara (approved):
Hej hej Hanna,

I really enjoy reviewing your codes. You made your code so clean and clear. I learned from the way you organize your files, the conditional css class, and was very impressed with your media query for various screen sizes.
You have completed all requirements of the Blue level, which I think you should be proud of. Even better, you managed to display the number of characters, change the text area into red, and also add validation to the submit button. Red level, here you go! :-). 
I put some suggestions on how to add a background image, how to throw and catch errors, for your reference.
Happy coding Hanna!!!

Tara

code/src/App.js
```diff
@@ -1,9 +1,66 @@
...
+  const fetchMessages = () => {
+    fetch(MESSAGES_URL)
+      .then(response => {
+        return response.json();
+      })
+      //Set the state (data is an array of messeges):
+      .then(data => {
+        //Filter out messages that don't have any text:
+        const filteredData = data.filter(message => {
+          return message.message;
```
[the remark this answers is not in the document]
EmOlsz: 👍

PR #123, opened 2020-11-01
Project Happy Thoughts - Amanda

amandaflod:
Link to my deployed project: 

https://pensive-shaw-14b90b.netlify.app/

I think it turned out well! I was struggling a lot at the beginning to get a grasp around useEffect (still not 100% sure I know what I've done 😅) I tried to implement the given style as much as possible, which was a fun thing to practice.


code/src/Components/happyThought.js
```diff
@@ -0,0 +1,54 @@
...
+    })
+      .then(() => {
+        window.location.reload()
+      })
+  }
+
+  return (
+    <div>
+      {thoughts && thoughts.map((thought) => (
+        <React.Fragment key={thought._id}>
```
liza-minican, line 31: This is an interesting usage - React.Fragment, will look into it to learn more:)

liza-minican (approved):
I enjoyed reviewing your code and didn't have much to comment on actually:) Good job! Wanted to mention a small detail - when the message is too long it goes outside the box. You could try to add `overflow-wrap: break-word` to the happy thought css. Congratulations for completing the Blue level and a part of Red too!!

PR #125, opened 2020-11-01
Happy Thoughts - Annika Melin

ankimelin:
Link to my deployed project: 
https://laughing-haibt-fe11b3.netlify.app/

Happy after this supertough week! 

Started off having to follow each step of Van's Monday code session. In the end (after a lot of googling, team discussions and checking other's code) I quite freely tried out different approaches on where to put the API fetches and how to pass props.

The importance of reading API documentation and where to find that.


code/src/App.js
```diff
@@ -1,9 +1,29 @@
-import React from 'react'
+import React, { useState } from 'react'
```
amandaflod, line 1: not really anything to comment anything on here, it works well and is readable, good job! 👏

code/src/index.css
```diff
@@ -1,13 +1,88 @@
...
+  display: flex;
+  height: 20px;
+}
+
+.likes {
+  display: inline-block;
+  margin-left: 10px;
+}
+
+/*** Media queries ***/
```
amandaflod, line 68: For this specific project, I would almost recommend removing the media queries and instead adding a max-width and fixed margin to the card,  it makes the transitions between different screensizes smoother! The padding on the body I think looks great on all screensizes 😊

code/src/index.css
```diff
@@ -1,13 +1,88 @@
...
-  -webkit-font-smoothing: antialiased;
-  -moz-osx-font-smoothing: grayscale;
+  background-color: rgb(255, 144, 162);
+  font-family: 'Roboto', sans-serif;
 }
 
-code {
-  font-family: source-code-pro, Menlo, Monaco, Consolas, "Courier New",
-    monospace;
+.card {
```
amandaflod, line 7: Someone had added a suuuuperlong word as a message just as I was looking at your code, and then I realized that it spilled over on your card (It did on my page too, so I have to fix that as well 😝) 
Apparently, you can add 
word-wrap: break-word;
to the card to make it break up the words that don't fit!

code/src/index.css
```diff
@@ -1,13 +1,88 @@
...
+
+.text-input {
+  padding: 5px;
+  width: calc(100% - 10px);
+  height: 50px;
+  background-color: rgb(243, 243, 243);
+  font-size: 12px;
+}
+
+.submit-button {
```
amandaflod, line 30: Super tiny thinhg, but it could be nice with some :hover and/or :active effects on the button!

code/src/index.css
```diff
@@ -1,13 +1,88 @@
...
+  justify-content: space-between;
+  color: grey;
+  font-size: 12px;
+}
+
+.likes-container {
+  display: flex;
+}
+
+button {
```
amandaflod, line 51: same thing on the heart! ❤️

amandaflod (changes requested):
***Changes requested in CSS***
Super great work on the project, all the JSX code I thought was very readable and everything was working as it should! 🎉 I have a few comments on the CSS as I discovered a problem (which I also had in my own code) when someone had added a message with a single superlong word, so have a look at that and update it, it's a super-quick fix! Also, the transitions between screensizes could be improves by removing th media queries and adding some other stuff instead, more on that below! 🤓
You are super awesome Annika! 🐯🚀🥇

code/src/components/Message.js
```diff
@@ -0,0 +1,29 @@
...
+        getMessages()
+      })
+  }
+
+  return (
+    <div className="card">
+      <p>{message} </p>
+      <div className="card-footer">
+        <div className="likes-container">
+          <button onClick={postLike}><img src="https://cdn4.iconfinder.com/data/icons/basic-ui-2-line/32/heart-love-like-likes-loved-favorite-256.png" alt="heart"></img></button>
```
EmOlsz, line 22: Images are self-closing tags

code/src/components/MessageInput.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const MESSAGE_URL = 'https://happy-thoughts-technigo.herokuapp.com/thoughts'
+
+  const [message, setMessage] = useState("")
+
+  const postMessage = message => {
+    fetch(MESSAGE_URL, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ message: message })
```
EmOlsz, line 13: Since both key and value inside this object have same name, we can simplify this :

```
```

code/src/index.css
ankimelin: Such a great tip, thanks Amanda!! Updated now when revisiting the project :-)

PR #129, opened 2020-11-01
Evelinabe - Happy thoughts

evelinabe:
Link to my deployed project: https://happiest-happy-thoughts.netlify.app/

I'm still lagging behind. The React part of the blue level is done, I will continue on with the styling.


code/src/App.js
```diff
@@ -1,9 +1,37 @@
...
+  const [thoughts, setThoughts] = useState([]);
+
+  const fetchThoughts = () => {
+    fetch(THOUGHTS_URL)
+      .then((res) => res.json())
+      .then((thought) => setThoughts(thought))
+      .catch((error) => console.error(error));
+  }
+
+  const postThoughts = (newThought) => {
```
mats-johansen, line 17: You don't need () when passing only one argument here. On the other hand the code almost look's better with (newThought) then newThought :D

code/src/components/Thought.js
```diff
@@ -0,0 +1,23 @@
...
+        {message}
+        <span className="thought-time">{time}</span>
+        <span className="liked-container">
+          <ThoughtLikes id={id} onLikedThought={onLikedThought} />
+          <span className="liked"> x {hearts}</span>
+        </span>
+      </p>
+    </div>
+  );
+};
```
mats-johansen, line 23: Very clean code in all the components!

code/src/components/ThoughtLikes.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export const ThoughtLikes = ({ id, onLikedThought }) => {
+  return (
+    <button type="button" onClick={() => onLikedThought(id)}>
+      <span role="img" aria-label="red heart">❤️</span>
+    </button>
+  )
```
mats-johansen, line 8: ;

mats-johansen (approved):
Great job! It look's look's and works great. I see no problems with the code or the functionality besides when writing a long message without any whitespace, like this:
ddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddda

It displays this outside the div. It's best to try it for yourself because of my bad explanation . I think adding word-wrap:break-word; in your css will do the job. 

Once again, great job!

code/src/components/Thought.js
```diff
@@ -0,0 +1,23 @@
...
+export const Thought = ({
+  message,
+  hearts,
+  time,
+  id,
+  onLikedThought
+}) => {
+  return (
+    <div className="thought-container">
+      <p key={id} id={id}>
```
EmOlsz, line 13: You don't need to use `key` prop here, we are not performing any loop

code/src/App.js
evelinabe: Tried to remove it but then it wouldn't compile :(

code/src/components/Thought.js
evelinabe: Removed.

PR #131, opened 2020-11-02
Project Happy Thoughts

ssundblom:
Link to my deployed project:


ankimelin (approved):
Great work Sara with last week's project!! Easy to follow code both in js and css files, no clutter just code that is down to the point - I really like your code style, and that you probably have spent some time on cleaning it up. 

Great choise to have separated css files corresponding to each component, makes it easy to understand what belongs where. 

So cool you implemented the character count - and it looks so easy when you do it! Also you managed to change the heart color depending on likes. 

Fantastic work and a supercute styling, of course!

code/src/App.js
```diff
@@ -1,9 +1,15 @@
 import React from 'react'
+import { MessageList } from 'MessageList'
+import { MessageInput } from 'MessageInput'
```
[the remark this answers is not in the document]
EmOlsz: Even if we have only few imports, we should stick to those rules

code/src/MessageInput.js
```diff
@@ -0,0 +1,42 @@
...
+    const handleSubmit = event => {
+        event.preventDefault()
+
+        fetch(MESSAGES_URL,
+            {
+                method: "POST",
+                headers: {
+                    'Content-Type': 'application/json'
+                },
+                body: JSON.stringify({ message: message })
```
EmOlsz, line 18: Or we can simplify this (since both left and right side of object have same name) :

```
```

code/src/MessageList.js
```diff
@@ -0,0 +1,31 @@
...
+                })
+                setMessages(filteredData)
+            })
+    }, [])
+
+    return (
+        <div>
+            {
+                messages.map(message => {
+                    return <HappyThought key={message._id} props={message} />
```
EmOlsz, line 26: That's not the best idea to name our prop `props`, it doesn't tell us anything about stored value. Far better would for example be :

```
return <HappyThought key={message._id} singleThought={message} />
```

PR #133, opened 2020-11-03
happy thoughts app

destinyleia:
https://happy-thoughts-happy-life.netlify.app/

Nothing works, I gave up after 6 h hours trying to solve it and completely change everything but still didn't work. So didn't update the new code I did (cause I didn’t work anyway) NEED HELP! 😄

Why nothing works together ... tried in my new code to do a fetch in the App-component instead and my props in the other components but that still didn't work...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,10 @@
 # Happy Thoughts
-
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+This week we were suppose to do likw a twitter app but ofr happy thoughts with React.  
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+Unfortunately I did not solve the problem with how to update the api with happy thoughts from my app or get the likes part to work. I have had a very stressful week so hav'nt had the time to really understand how to solve the problem. I had all the design in. But the code does not work as it should. Wish I had the time to really solve it and understand it. 
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+
--- a/README.md
+++ b/README.md
@@ -6,5 +6,5 @@ This week we were suppose to do likw a twitter app but ofr happy thoughts with R
 Unfortunately I did not solve the problem with how to update the api with happy thoughts from my app or get the likes part to work. I have had a very stressful week so hav'nt had the time to really understand how to solve the problem. I had all the design in. But the code does not work as it should. Wish I had the time to really solve it and understand it. 
 
 ## View it live
-
+https://happy-thoughts-happy-life.netlify.app/
 
--- a/code/package.json
+++ b/code/package.json
@@ -10,6 +10,7 @@
     "eslint-plugin-jsx-a11y": "^6.3.1",
     "eslint-plugin-react": "^7.21.3",
     "eslint-plugin-react-hooks": "^4.1.2",
+    "moment": "^2.29.1",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "react-scripts": "3.4.3"
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,10 @@
 import React from 'react'
+import { MessageList } from './components/MessageList'
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <>
+    <MessageList />
+    </>
   )
 }
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,12 @@
 import React from 'react'
+
 import { MessageList } from './components/MessageList'
+import { NewThought } from './components/NewThought'
 
 export const App = () => {
   return (
     <>
+    <NewThought />
     <MessageList />
     </>
   )
--- a/code/src/components/MessageList.js
+++ b/code/src/components/MessageList.js
@@ -1,0 +1,38 @@
+import React, { useEffect, useState } from 'react'
+import moment from 'moment'
+
+export const MessageList = () => {
+  const MESSAGES_URL = 'https://wk11livesession.herokuapp.com/messages'
+  const [messages, setMessages] = useState([]);
+
+  useEffect(() => {
+    fetch(MESSAGES_URL)
+      .then((res) => {
+        return res.json();
+      })
+      .then((data) => {
+        console.log(data)
+        data.reverse();
+
+        const filtredMessages = data.filter((message) => message.text)
+
+        setMessages(filtredMessages)
+      })
+  }, [])
+
+  return (
+    <div>
+      {messages.map((message) => {
+        return (
+          <p className="message" key={message._id}>
+            {message.text} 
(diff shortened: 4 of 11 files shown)


README.md
```diff
@@ -1,13 +1,10 @@
 # Happy Thoughts
```
VanessaSue27, line 1: I think once you come back to this project and make it work, would be nice to update the ReadMe files, so they look complete if you would like to include this project in your portfolio ✨

code/package.json
```diff
@@ -10,6 +10,7 @@
     "eslint-plugin-jsx-a11y": "^6.3.1",
```
VanessaSue27, line 10: There seems to be some problem when you are doing the npm install, because I see you have at least 3 different package-lock.json and 2 different package.json, when you should only have one of each I think? 👀
Make sure to be in code folder when doing the npm install.

code/src/App.js
```diff
@@ -1,9 +1,13 @@
 import React from 'react'
 
+import { MessageList } from './components/MessageList'
+import { NewThought } from './components/NewThought'
```
VanessaSue27, line 4: Nice keeping a good order and spacing of all the imported items 👍

code/src/App.js
```diff
@@ -1,9 +1,13 @@
...
+
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <>
+    <NewThought />
+    <MessageList />
```
VanessaSue27, line 10: Indentation here, these two components should be two spaces in.

code/src/components/MessageList.js
```diff
@@ -0,0 +1,60 @@
+import React, { useEffect, useState } from 'react'
+
+import moment from 'moment'
+
+export const MessageList = () => {
+  const MESSAGES_URL = 'https://wk11livesession.herokuapp.com/messages'
```
VanessaSue27, line 6: You are fetching the messages from the wrong URL 😅. This URL is the one we used during the lessons, but the one for the project should be:
https://happy-thoughts-technigo.herokuapp.com/thoughts 💕

code/src/components/MessageList.js
```diff
@@ -0,0 +1,60 @@
...
+
+    const fetchMessages = () => {
+    fetch(MESSAGES_URL)
+      .then((res) => {
+        return res.json();
+      })
+      .then((data) => {
+        data.reverse();
+
+        const filtredMessages = data.filter((message) => message.text)
```
VanessaSue27, line 21: Great filtering for empty messages! 😎

code/src/components/MessageList.js
```diff
@@ -0,0 +1,60 @@
...
+                <span className="message-time">
+                  {moment(message.created).fromNow()}
+                </span>
+              </p>
+              </div>
+              )
+      })}
+          </>
+        )
+      }
```
VanessaSue27, line 60: Careful with indentation here, maybe installing a code formatter could be of help? 👩‍💻

code/src/components/MessageList.js
```diff
@@ -0,0 +1,60 @@
...
+        data.reverse();
+
+        const filtredMessages = data.filter((message) => message.text)
+
+        setMessages(filtredMessages)
+      })
+}
+
+  const likeHearts = (messageId) => {
+    fetch(`https://happy-thoughts-technigo.herokuapp.com/thoughts/${messageId}/like`, {
```
VanessaSue27, line 28: There is a POST 404 Bad Request error coming up when clicking on Heart button 💔
You have followed the right steps to implement this, the problem is that since you fetched messages from the wrong API, the data you got is not the correct one for this project. So you are not getting the correct message IDs.
In the console you'll see that this is the URL we are trying to send the POST request to in order to like a thought. We get error 404 because the URL is wrong.
It looks like this:
**https://happy-thoughts-technigo.herokuapp.com/thoughts/[object%20Object]/like**
Notice the place where there should be a messageID, we get this instead: **[object%20Object]** - so the POST does not go through.
Maybe try doing the fetch to the correct URL and it will probaby work as it should 💥

code/src/components/NewThought.js
```diff
@@ -0,0 +1,39 @@
+import React, { useState } from 'react'
+
```
VanessaSue27, line 2: Check indentation for this component, I think you might have your tab set to 4 spaces instead of 2. 
Installing a formatter could be of help, I'm using Prettier sometimes ✨

code/src/components/NewThought.js
```diff
@@ -0,0 +1,39 @@
...
+export const NewThought = ({ onChangeThought }) => {
```
VanessaSue27, line 3: There is no onChangeThought function defined anywhere. It seems you wanted to pass it in here as a prop, but it does not exist in App.js either, so even though you call this function later on, nothing happens because the function doesn't exist.
Maybe try writing the function in App and passing it as a prop, or writing it directly in this component 💪

code/src/components/NewThought.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    const handleMessages = e => {
+        setNewThought(e.target.value)
+    }
+
+    return (
+        <div className="new-thought-section">
+            <form onSubmit={handleSubmit}>
+                <label for="title-thought-section">What's making you happy right now?</label>
```
VanessaSue27, line 19: I think here if you want the label to be connected to the textarea then the label's "for" attribute should be the same as the textarea's "id" attribute, that way they are connected. 
Or this could be a <p> element and no "for" is needed.

code/src/components/NewThought.js
```diff
@@ -0,0 +1,39 @@
...
+                    id="newHappyThought"
+                    name="newHappyThought"
+                    rows="5"
+                    cols="36"
+                    value={newThought}
+                    onChange={handleMessages}
+                    required
+                />
+            <br></br>
+            <button type="button" value="Submit">
```
VanessaSue27, line 30: No functionality has been implemented to this button. In order for this button to do the submit for the form, you need to change the type to type="submit". That way the function handleSubmit gets called when the button is clicked 🙃

You are also missing the POST request in which we post a new thought to the API. This should probably be part of the onChangeThought function that is missing. The idea is that when you click on the button, we call the handleSubmit function which will then do a new fetch with method POST and add this new thought to the API 💌

code/src/components/NewThought.js
```diff
@@ -0,0 +1,39 @@
...
+export const NewThought = ({ onChangeThought }) => {
+    const [newThought, setNewThought] = useState("")
+
+    const handleSubmit = e => {
+        e.preventDefault()
+        onChangeThought(newThought)
+        setNewThought('')
+    }
+
+    const handleMessages = e => {
```
VanessaSue27, line 12: 💯

VanessaSue27 (changes requested):
Hello Destiny!
I think you are SO close to getting this to work. Reviewing your code, I noticed some errors that are the ones preventing it to work.
These are the main things I noticed that need to be addressed (and I talk more specifically about them on my comments):
- Fetch to the correct API (use the right URL)
- Define the onChangeThought function which includes the POST request to add a new thought.
- Overall indentation.

You got all the CSS in place, so not much is missing, you can do this! 💪💌
Hope my review helps getting you to the finish line! 😎
//Vanessa.

PR #137, opened 2020-11-05
Happy thoughts app 

destinyleia:

https://happy-thoughts-happy-life.netlify.app/

See my read me! 😍 Vanessa did a fantastic code review on my code and helped me out! Unfortunately it disappeared when I updated my new code.. don't know why... but she did a great job reviewing my code and helped alot! 😁

How it all works together, with props , useEffect, state etc...

### DIFF
--- a/code/src/components/MessageList.js
+++ b/code/src/components/MessageList.js
@@ -1,0 +1,69 @@
+import React, { useEffect, useState } from 'react'
+
+import moment from 'moment'
+
+export const MessageList = () => {
+    const MESSAGES_URL = 'https://happy-thoughts-technigo.herokuapp.com/thoughts'
+    const [messages, setMessages] = useState([]);
+
+
+    useEffect(() => {
+        fetchMessages()
+    }, [])
+
+    const fetchMessages = () => {
+        fetch(MESSAGES_URL)
+            .then((res) => {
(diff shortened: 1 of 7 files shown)


code/src/components/MessageList.js
```diff
@@ -0,0 +1,69 @@
...
+
+    return (
+        <>
+            <div className="message-list">
+                {messages.map((message) => {
+                    return (
+                        <div key={message._id} className="message-section">
+                            <p className="message" key={message._id}>
+                                {message.message}
+                                <br></br>
```
EmOlsz, line 44: We shouldn't use `<br />` tags, they are deprecated. You should use proper styles in CSS file in order to achieve same result. Maybe `display` property will help?

code/src/components/MessageList.js
```diff
@@ -0,0 +1,69 @@
...
+                            </p>
+                            <div className="heartbutton-section">
+                                <div className="heartbutton-x-section">
+                                    <button
+                                        className="likeheart-button"
+                                        id={`hearts-button${message._id}`}
+                                        key={message._id}
```
EmOlsz, line 51: We don't need to add `key` attribute anywhere else apart from root loop element (in your case it's `message-section` element)

PR #138, opened 2020-11-11
Agnes happy journal

aagneshansson:
Link to my deployed project:
https://stayhappy.netlify.app/

### DIFF
--- a/code/src/HappyForm.css
+++ b/code/src/HappyForm.css
@@ -12,7 +12,7 @@
 
 .happy-form textarea {
     width: 100%;
-    resize: none; 
+    resize: none;
     padding: 10px;
     font-size: 16px;
     margin: 10px 0;
@@ -53,4 +53,3 @@
         width: 500px;
     }
 }
-
--- a/code/src/HappyForm.css
+++ b/code/src/HappyForm.css
@@ -1,0 +1,46 @@
+.happy-form {
+    background-color: #f3f1f1;
+    border: 2px solid black;
+    box-shadow: 4px 4px black;
+    width: 300px;
+    margin: 16px;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    padding: 20px;
+}
+
+.happy-form textarea {
+    width: 100%;
+    resize: none; 
+    padding: 10px;
+    font-size: 16px;
+    box-sizing: border-box;
+    margin: 10px 0;
+}
+
+.happy-form h3 {
+    margin: 0;
+}
+
+.happy-form button {
+    height: 50px;
+    width: 60%;
+    background-color: #ffadad;
+    border: none;
+    font-size: 14px;
+    border-radius: 25px;
+    padding: 10px;
+    font-weight: 500;
+}
+
+.happy-form button:hover {
+    cursor: pointer;
+}
+
+.form-footer {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    width: 100%;
+}
--- a/code/src/HappyForm.css
+++ b/code/src/HappyForm.css
@@ -1,21 +1,13 @@
-* {
-    -webkit-box-sizing: border-box;
-    -moz-box-sizing: border-box;
-    box-sizing: border-box;
-    -webkit-font-smoothing: antialiased;
-    text-rendering: optimizeLegibility;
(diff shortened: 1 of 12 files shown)


code/src/HappyForm.css
```diff
@@ -0,0 +1,55 @@
+.happy-form {
+    width: 300px;
+    margin: 16px;
+    padding: 20px;
+    background-color: #f5f5f1;
+    border: 1px solid black;
```
evelinabe, line 6: Just for consistency you could write the color as `#000` instead of `black`

code/src/HappyForm.js
```diff
@@ -0,0 +1,58 @@
...
+                valure={message}
+                onChange={event => setMessage(event.target.value)}
+            >
+            </textarea>
+
+            <div className="form-footer">
+                <button
+                    type='submit'
+                    onClick={handleSubmit}
+                    disabled={message.lenght < 6 || message.lenght > 140 ? true : false}
```
evelinabe, line 43: Spelling mistake here - it should read `.length`.

code/src/HappyForm.js
```diff
@@ -0,0 +1,58 @@
...
+                >
+                    <span role="img" aria-label="Red heart"> &#10084;&#65039;</span>
+
+                        Send a happy thought!
+                    </button>
+                <p className="text-validation">
+                    <span style={{ color: message.lenght < 6 || message.lenght > 140 ? "red" : "#000" }}>
```
evelinabe, line 50: Spelling mistake on `.length` here as well.

code/src/HappyForm.css
```diff
@@ -0,0 +1,55 @@
...
+    font-size: 16px;
+    border-radius: 20px;
+    padding: 10px 20px;
+    font-weight: 500;
+    border: 1px solid #f5f5f1;
+}
+
+.happy-form button:hover {
+    cursor: pointer;
+    background: white;
```
evelinabe, line 40: For consistency, color should be `#fff`.

code/src/HappyForm.js
```diff
@@ -0,0 +1,58 @@
...
+        })
+            .catch(err => console.log("error:", err))
+    }
+
+    return (
+        <form className='happy-form'>
+            <h3>Post a happy thought!</h3>
+            <textarea
+                rows='3'
+                valure={message}
```
evelinabe, line 34: Should it be `value`?

evelinabe (changes requested):
Great job with your happy journal, Agnes! 
I like the structure with components and separate CSS-files. The comments are nice, and just enough to make the code more readable. 

There are two things I'd like you to change. First, your function for disabling button doesn't work as intended due to a spelling mistake. The same goes for the span which should show different colors when the message has the right length. Fix the spelling and it should be good to go!

The second thing is that the text won't break if you write a really, really, really long word. This could be fixed by adding `word-wrap: break-word` in the styling.

code/src/HappyForm.js
```diff
@@ -0,0 +1,58 @@
+import React, { useState } from 'react'
+import './HappyForm.css'
+
+const Happyform = "https://happy-thoughts-technigo.herokuapp.com/thoughts";
```
EmOlsz, line 4: This is not the best name for url variable :

1. First of all, only components should have names written in uppercase
2. Something like `MESSAGE_URL` would be far better

code/src/HappyForm.js
```diff
@@ -0,0 +1,58 @@
...
+        event.preventDefault();
+
+        // Sending a POST request with the message state 
+        fetch(Happyform,
+            {
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json'
+                },
+                body: JSON.stringify({ message: message }),
```
EmOlsz, line 19: Since both left and right side of object has same name, we can simplify it :

```
```

code/src/HappyForm.js
```diff
@@ -0,0 +1,58 @@
...
+            }
+        ).then(() => {
+            setMessage('')
+            // props.onFormSubmit(message)
+            window.location.reload();
+        })
+            .catch(err => console.log("error:", err))
```
EmOlsz, line 26: `console.error` would be more appropriate

PR #152, opened 2021-03-26
Happy Thoughts project. Vladimir Bykov

vladjnbykov:
Link to my deployed project: https://elated-thompson-c0e3cd.netlify.app/

My reflections on how this week's project turned out: was OK

Things I'd like to have clarified or explained in more detail: More explanation of React data transfer (probs and states ) is necessary.

### DIFF
--- a/code/src/components/Form.js
+++ b/code/src/components/Form.js
@@ -46,22 +46,26 @@ export const Form = () => {
 
     return (
         <div>
-            <form onSubmit={onFormSubmit}>
-                <label htmlFor="newMessage">Write new message!</label>
-                <input
+            <form className="form" onSubmit={onFormSubmit}>
+                <label htmlFor="newMessage">What's make you happy right now?</label>
+                <input className="input"
                     id="newMessage"
                     type="text"
                     value={messageNew}
                     onChange={onMessageNewChange}
                 />    
-                <button type="submit">Send message!</button>
+                <button className="btn-submit" type="submit">💗 Send Happy Thought 💗</button>
 
             </form>
 
             {messageList.map(message => (
-                <div key={message._id}>
+                <div className="message-body" key={message._id}>
                     <h4>{message.message}</h4>
-                    <p className="time-from-now">posted {moment(message.createdAt).fromNow()}</p>
+                    <div className="like">
+                        <button className="btn-like">💗</button>
+                        <p>x {message.hearts}</p>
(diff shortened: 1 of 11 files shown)


code/src/components/Form.js
```diff
@@ -0,0 +1,90 @@
...
+
+  const handleFormSubmit = (event) => {
+    event.preventDefault()
+
+    const options = {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({ message: messageNew })
```
Ingetora, line 36: Nice usage!

code/src/components/MessageElement.js
```diff
@@ -0,0 +1,20 @@
...
+import React from 'react'
+import moment from 'moment'
+
+const MessageElement = ({ message, onLikesIncrease }) => {
+    return (
+        <div className="message-body">
+            <h4>{message.message}</h4>
+            <div className="like">
+                <button onClick={() => onLikesIncrease(message._id)} className={message.hearts === 0 ? "btn-start": "btn-like"}>💗</button>
+                <p>x {message.hearts}</p>
```
Ingetora, line 12: Mabye it's possible to shorten this code, or use different intendentions?

code/src/index.css
```diff
@@ -1,13 +1,93 @@
...
+
+.btn-like {
+  background-color: pink;
+  border: none;
+  border-radius: 40px;
+  height: 40px;
+  width: 40px;
+  margin-right: 10px;
+  outline: none;
+}
```
Ingetora, line 93: I would like it if your site showed responsiveness in different size screens.

Ingetora (commented):
Very nice looking project! It's evident that you have worked hard on your code. Terrific job, Vlad!

KarolCislowski (changes requested):
Hej!
Great job, but please clean your code from all console.log. 🧹 

About props and states:
The whole idea of React is to make your application react to the events by ultrafast re-renderings of page content. But to make it efficient react needs to know which operations can change the DOM and which 100% not. So it performs checking only if one of two things happens: there was a change of state or props. 
The state is the local storage of component. If we have a situation that the Component has a list of child components and every one of them needs a different part of the data we need to pass it. We passing them inside the props.

KarolCislowski (approved):
Great 🥇

PR #155, opened 2021-03-27
Project happy thoughts-Andrea

Andrealarsson:
Link to my deployed project:
https://happy-thought.netlify.app/

At first, I had some problems with how to structure the code. I thought I could start writing in the components and then just collect them in app.js. But then I learned that I should always write the code in app.js and then divide some of the code into different components. But overall this week went well.

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -4,7 +4,7 @@ import React, { useState, useEffect } from 'react'
 import ThoughtForm from './components/ThoughtForm'
 import ThoughtList from './components/ThoughtList'
 
-import { API_URL, HEARTS_URL } from 'reusable/urls'
+import { API_URL, HEARTS_URL } from './reusable/urls'
 
 export const App = () => {
   const [thoughtList, setThoughtList] = useState([])
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -4,7 +4,7 @@ import React, { useState, useEffect } from 'react'
 import ThoughtForm from './components/ThoughtForm'
 import ThoughtList from './components/ThoughtList'
 
-import { API_URL, HEARTS_URL } from './reusable/urls'
+import { API_URL, HEARTS_URL } from './reusable/Urls'
 
 export const App = () => {
   const [thoughtList, setThoughtList] = useState([])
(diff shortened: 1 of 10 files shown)


code/src/components/ThoughtForm.js
```diff
@@ -0,0 +1,36 @@
...
+            disabled={
+            thoughtNew.length < 5 || thoughtNew.length > 140 ? true : false}>
+            <span role="img" aria-label="heart">❤️</span> Send Happy Thought <span role="img" aria-label="heart">❤️</span>
+          </button>
+          <p className="characters-counter">
+            <span
+              style={{ color: thoughtNew.length >= 140 ? '#ff3933' : '#000' }}>
+              {thoughtNew.length}
+            </span>
+            /140
```
Aepera, line 28: This is very nice, counting the characters entered. Also the submit button changing color when a message has been entered 🎉

Aepera (approved):
This is a very nice web app that is working exactly like supposed to. Good job with the design, it is quite similar to the guide. The site is responsive, however not centered on mobile view.

The code is clean. short and easy to follow. It is nice that the JSX has been divided into three components. I like especially the character counter next to the text field! 👍

One suggestion to add, would be hover effects on the buttons. 👍

code/src/App.js
```diff
@@ -1,9 +1,88 @@
...
+      .catch(err => console.error(err))
+  }
+
+  //FETCH TO POST NEW MESSAGE
+  const handleThoughtNewChanged = (event) => {
+  setThoughtNew(event.target.value)
+  }
+
+  const reloadPage = () => {
+    window.location.reload()
```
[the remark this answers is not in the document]
KarolCislowski: It is an anti-pattern to reload single page application. You need to rethink the state/props structure to make your app re-render

code/src/App.js
```diff
@@ -1,9 +1,88 @@
...
+  useEffect(() => {
+    fetchThoughtList()
+  }, [])
+
+  //FETCH TO IMPORT THOUGHTS
+  const fetchThoughtList = () => {
+    fetch (API_URL)
+      .then(response => response.json())
+      .then(thought => setThoughtList(thought))
+      .catch(err => console.error(err))
```
KarolCislowski, line 22: You shouldn't print anything to console in production code. Always remove functions like this after debugging.

KarolCislowski (changes requested):
Good code and nice styling, but please change your logic to avoid refreshing the page and remove all console outputs.

KarolCislowski (approved):

PR #162, opened 2021-03-28
Week 11 - Project Happy Thoughts - Julia Ros

julialeonieros:
Link to my deployed project:
https://determined-engelbart-a7f7f6.netlify.app/

I aimed for the app to resemble the image in the project brief, and I am quite happy with the result. If I would have had more time I'd probably add an animation. For now I only added a character counter.

I just realised I forgot to make the site responsive, and I'll try to do that in the near future.


code/src/App.js
```diff
@@ -1,9 +1,77 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
+
```
Ann-Sofi, line 2: Good that you left a empty line between the outer and local imports.

code/src/App.js
```diff
@@ -1,9 +1,77 @@
...
+  }
+
   return (
-    <div>
-      Find me in src/app.js!
+    <div className="main-container">
+      <Form 
+        onSubmit={handleFormSubmit}
+        value={messageNew}
+        onChange={handleMessageNewChange}
```
Ann-Sofi, line 68: Good naming with the "handle". Now I know that this is beeing forwarded as props to a component.

code/src/Components/Form.js
```diff
@@ -0,0 +1,36 @@
...
+              className="form-button"
+              >
+              <span role="img" aria-labelledby="Heart">❤️ </span>
+              <span className="button-text">Send happy thought!</span>
+              <span role="img" aria-labelledby="Heart"> ❤️</span>
+            </button>
+          </form>
+        </div>
+        <div>
+          <p className={value.length > 134 ? "character-counter red" : "character-counter black" }>Characters left: {(140-value.length)} 📝</p>
```
Ann-Sofi, line 30: Nice and short solution! :-)

Ann-Sofi (approved):
Cute site and it works as it should at the blue level. 

Good commit messages. I can follow your process by looking at them 👍

SInce you added a "warning" when the text is getting to long, I think the next thing to add could be to add a warning when it's to short. It could be pretty easy to reach the maximum of letters, but it might be just as common to write a very short one?

code/src/App.js
```diff
@@ -1,9 +1,77 @@
...
+      .then(messages => setMessageList(messages))
+      .catch(error => console.error(error))
+  }
+
+  const handleMessageNewChange = (event) => {
+    setMessageNew(event.target.value)
+  }
+
+  const refreshPage = () => {
+    window.location.reload()
```
KarolCislowski, line 27: It is an anti-pattern to reload single page application. You need to rethink the state/props structure to make your app re-render

code/src/App.js
```diff
@@ -1,9 +1,77 @@
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      }
+    }
+
+    fetch(LIKES_URL(id), options)
+      .then(result => result.json())
+      .then(() => fetchMessageList())
+      .catch(error => console.error(error))
```
KarolCislowski, line 60: console.log can't be present in the production code.

KarolCislowski (changes requested):
Hej!
Great job, but please:
1. clean your code from all console.log. 🧹 
2. find a way to avoid refreshing.
3. Make it responsive

julialeonieros:
Hi Karol!
I've now made it responsive and removed the refresh-function (instead set the state of messageNew to empty string). I also removed the console.errors (although Maks used console.errors like that in the lecture so I thought it was ok, but I guess you should remove them when you see that the code works?).

KarolCislowski (approved):
Awesome app!!

PR #168, opened 2021-03-28
Happy Thoughts - Tobias Hagberg, Team Tigers, class jan 2021

TobiasHagberg:
Link to my deployed project: https://epic-bassi-2d5a3d.netlify.app/

A lot of code were handed to us during lectures and that appoach allowed for more time on trying to understand the concept of useState and useEffect. I liked it. Fun project, nice design but unfortunaltely I was not able to keep functionalty when breaking out code into smaller components sp reverted back to having everything in App.js

I jsut need more examples of use cases and more practise :)

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -63,20 +63,21 @@ export const App = () => {
 
   return (
     <main>
-      <form onSubmit={onFormSubmit}>
+      <form className="thought-card" onSubmit={onFormSubmit}>
         <label htmlFor="newThought">Send a happy thought!</label>
         <input 
+          className="send-thought_input-field"
           id="newThought"
           type="text"
           value={thoughtNew}  
           onChange={onThoughtNewChange} 
         />
-        <button type="submit">Send thought!</button>
+        <button className="send-thought_button" type="submit">Send Happy Thought!</button>
       </form>
       {thoughtList.map(thought => (
-        <div key={thought._id}>
+        <div className="thought-card" key={thought._id}>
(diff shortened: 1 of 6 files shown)


sandra-wallen (commented):
**APPROVED**

Hi!

Nice project! The code and indentions are clean and easy to read. It would have been nice with some different components but I saw in your read me that you are aware of this and that you tried but couldn't quite get it to work 😊.

I love the hovering effect on the heart btn, it looks really nice!

I noticed a few things with the design, the font doesn't seem to be the one in the example design but I think that you probably did that on purpose and I think that's fine, we should be allowed to do some small customizations 😁. Another thing is that the heart btn is not changing color when there is more likes than 0, you can implement that with a ternary operator on the className if you want 😊. Other than that I think your design and project follows the guidelines very well!

Well done! 🤩

code/src/App.js
```diff
@@ -1,9 +1,122 @@
...
+
+  useEffect(() => {
+    fetchThoughtList()
+  }, [])
+  
+  const fetchThoughtList = () => {
+    fetch (API_URL)
+      .then(res => res.json())
+      .then(thoughts => setThoughtList(thoughts))
+      .catch(err => console.error(err))
```
KarolCislowski, line 19: console outputs can't be present in the production code.

KarolCislowski (changes requested):
Hej!

KarolCislowski (approved):

PR #174, opened 2021-03-28
Happy Thoughts - Erika A Porath

TheErry:
Link to my deployed project:
https://happy-nobel-59adad.netlify.app/

This time the workflow was very straight forward. I coded along with Maks lectures so the code doesn't differ so much from his, except that I'm using the API for the project and the code is changed to fit that. Otherwise it's mostly the names of the components that's changed.  The styling was quite easy this time too. I think the project went well.

I couldn't find the exact fonts that were used in the given design, but I hope the ones I have are ok.


code/src/App.js
```diff
@@ -1,9 +1,76 @@
...
+    const options = {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json"
+      },
+      body: JSON.stringify({ message: messageNew })
+    }
+    fetch(API_URL, options)
+    .then(res => res.json())
+    .then(recievedMessage => setMessageList([recievedMessage, ...messageList])) //Switched the position of recievedMessage and messageList to make the new messages come first in the list
```
johlofg, line 38: Good comment

code/src/index.css
```diff
@@ -11,3 +18,124 @@ code {
...
+}
+
+.send-button {
+  display: flex;
+  align-items: center;
+  height: auto;
+  border-top-left-radius: 17px;
+  border-bottom-left-radius: 17px;
+  border-top-right-radius: 17px;
+  border-bottom-right-radius: 17px;
```
johlofg, line 64: since you have the same radius (17px) on all corners, you can simply write **border-radius: 17px;** to save time and space :)

johlofg (approved):
Not that much to add. Good job! and nice clean code. Maybe add some describing comment in the functions, otherwise 👍

code/src/App.js
```diff
@@ -1,9 +1,76 @@
...
+
+  useEffect(() => {
+    fetchMessageList()
+  }, [])
+
+  const fetchMessageList = () => {
+    fetch(API_URL)
+      .then(res => res.json())
+      .then(messages => setMessageList(messages))
+      .catch(err => console.error(err))
```
KarolCislowski, line 20: console outputs can't be present in the production code.

KarolCislowski (changes requested):
Hej!

KarolCislowski (approved):
😸

PR #178, opened 2021-03-28
Happy thoughts - Astrid Persson

Aepera:
Link to my deployed project:
https://react-happy-thoughts.netlify.app/

One small detail: the heart emoji inside the button is not exactly centered.


code/src/App.js
```diff
@@ -1,9 +1,88 @@
...
+          message.hearts += 1
+        }
+        return message
+      }) 
+      setMessageList(updatedMessageList)
+    })
+    .catch(error => console.log(error))
+  }
+
+
```
rrpersson, line 70: Your code is very clean. It's easy to read and understand!

code/src/components/Form.js
```diff
@@ -0,0 +1,21 @@
...
+        value={newMessage}
+        onChange={onNewMessage}
+        className='form-input'
+      />
+      <button type='submit' className='form-btn'>SEND HAPPY THOUGHT</button>
+    </form>
+  )
+}
+
+export default Form
```
rrpersson, line 21: I think as a tip you could make is so that your site will refresh automatically when you press submit on the form! Good job of having such a clean component.

code/src/components/Message.js
```diff
@@ -0,0 +1,31 @@
...
+      {messageList.map(message => (
+        <div key={message._id} className='message'> 
+
+          <div>
+            <h3>{message.message}</h3>
+          </div>
+          <div className='btn-container'> 
+
+            <div className='btn-counter'>
+              <button className= { `heart-btn ${message.hearts === 0 ? "heart-btn-unliked" : "heart-btn-liked"}`}
```
rrpersson, line 17: Very good use of ternary operators! Nice!

code/src/components/Urls.js
```diff
@@ -0,0 +1,6 @@
+
+
+export const API_URL = 'https://happy-thoughts-technigo.herokuapp.com/thoughts'
+
+
+export const APIHEARTS_URL = (THOUGHT_ID) => `https://happy-thoughts-technigo.herokuapp.com/thoughts/${THOUGHT_ID}/like`
```
rrpersson, line 6: Good use of a URLs file.

rrpersson (approved):
Very well done Astrid. I think you did a great job with the page and the design, it's almost 100% like the picture we had to re-create. I think you are writing very clean code. You've made really great progress!

code/src/components/Form.js
KarolCislowski: Not refresh, but re-render the list with re-fetched data. There should be never such thing as refreshing the single page application.

code/src/App.js
```diff
@@ -1,9 +1,88 @@
...
+  useEffect(() => {
+    fetchMessageList()
+  }, [])
+
+
+  const fetchMessageList = () => { 
+    fetch(API_URL)
+      .then(response => response.json())
+      .then(messages => setMessageList(messages))
+      .catch(error => console.log(error))
```
KarolCislowski, line 24: console outputs can't be present in the production code.

KarolCislowski (changes requested):
Nice project but you need to make some fixes to reach blue level:
1. remove all console outputs.
2. your messages list needs to re-render itself with newly fetch data.
3. your text field needs to be cleared after sending the info to the API.

KarolCislowski (approved):
👍🏻

PR #192, opened 2021-03-28
Happy-Thoughts-SandraWallén

sandra-wallen:
https://frosty-heyrovsky-0bfc89.netlify.app/

Building the Happy Thoughts project went pretty good, I started writing all the code in the App component and then devided things to different components. I didn't encounter any big problems but I am uncertain about how to think when deviding code in different components. I instinctly want to move functions to the components where they are being used, I feel that that makes the code more clean and readable. But I have noticed that this is causing trouble sometimes. As an example, I wanted to do a new GET fetch inside the POST fetch, but the fetches were in different components and not related to each other in a way that I could pass the fetch over as props. To handle this I made a copy of the function inside the other component, which I'm sure is not a good way to do it.

### DIFF
--- a/code/src/components/Header.js
+++ b/code/src/components/Header.js
@@ -1,0 +1,14 @@
+import React from 'react'
+
+const Header = () => {
+
+  return (
+    <article className="message-container header-container">
+      <p>
+        <span className="heart-letter-emoji" role="img" aria-label="heart-letter-emoji">💌</span>
+      </p>
+      <h2 className="header-container-heading">Happy Thoughts</h2>
+    </article>
+  )
+}
+export default Header
--- a/code/src/components/LikePost.js
+++ b/code/src/components/LikePost.js
@@ -1,0 +1,47 @@
+import React from 'react'
+
+import { LIKE_URL } from '../reusable/urls'
+
+const LikePost = ({ name, hearts, messageList, setMessageList, fetchMessageList }) => {
+
+  const handleClick = () => {
+    postHearts(name)
+  }
+
+  const postHearts = (messageId) => {
+    fetch(LIKE_URL(messageId), {
+      method: 'POST',
+      headers: { "Content-Type": "application/json" }
+    }) .then (() => {
+      onLiked(messageId)
+      fetchMessageList()
+    }) .catch(err => console.error(err))
+  }
+
+  const onLiked = (messageId) => {
+    const updatedMessageList = messageList.map(thought => {
+      if (thought._id === messageId) {
+        thought.hearts += 1
+      }
+      return thought
+    })
+    setMessageList(updatedMessageList)
+  }
+
+  return (
+    <p>
+      <button 
+      name={name} 
+      className={hearts === 0 ? 'like-btn zero' : 'like-btn'}
+      onClick={handleClick}>
+        <span 
+        className="heart-emoji" 
+        role="img" 
+        aria-label="heart-emoji">❤️
+        </span>
+      </button>
+      x {hearts}
+    </p>
+  )
+}
+export default LikePost
--- a/code/src/components/LikePost.js
+++ b/code/src/components/LikePost.js
@@ -9,13 +9,15 @@ const LikePost = ({ name, hearts, messageList, setMessageList, fetchMessageList 
   }
 
   const postHearts = (messageId) => {
-    fetch(LIKE_URL(messageId), {
+    fetch(LIKE_URL(messageId),{
       method: 'POST',
       headers: { "Content-Type": "application/json" }
-    }) .then (() => {
+    }) 
+      .then (() => {
       onLiked(messageId)
       fetchMessageList()
-    }) .catch(err => console.error(err))
+    }) 
+      .catch(err => console.error(err))
   }
 
   const onLiked = (messageId) => {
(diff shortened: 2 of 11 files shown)


code/src/components/Header.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+const Header = () => {
```
IsabellaM5, line 3: I would consider this component for a bit and the chosen element that you return here. In general, an article tag is usually intended for forum posts, newspaper articles, blog entries etc. Since you’re only displaying a heading and a span in this component, my suggestion would be to wrap this in a div tag. Another suggestion is to wrap it in a header tag, which you in App.js would have to mount outside of the main element that you currently have wrapping the PostMessage and ListMessage components.

Also, personally I would’ve expected to see a header element based on the naming of the component, but that's just me maybe 🙂

code/src/components/PostMessage.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react'
+
+import { HAPPY_THOUGHTS_API } from '../reusable/urls'
+
+const PostMessage = ({ messageNew, setMessageNew, messageList, setMessageList, chars, setChars }) => {
```
IsabellaM5, line 5: messageList - do you need this in this component? I don't see you using it hehe 🧐🙂

code/src/components/PostMessage.js
```diff
@@ -0,0 +1,57 @@
...
+      },
+      body: JSON.stringify({ message: messageNew })
+    }
+
+    fetch(HAPPY_THOUGHTS_API, postArgument)
+      .then(response => response.json())
+      .then(() => {
+        setMessageNew('')
+        fetchMessageList()})
+  }
```
IsabellaM5, line 35: One suggestion is to add a .catch here like in the fetch above when you fetch messages. 🙂

code/src/components/PostMessage.js
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <article className="message-container form-container">
```
IsabellaM5, line 38: I would consider the usage of an article tag here. As explained in my comment on Header.js, article tags are generally intended for forum posts, newspaper articles, blog entries etc. If you need a wrapper here I would consider using a section element instead 🙂

code/src/components/PostMessage.js
```diff
@@ -0,0 +1,57 @@
...
+      <form className="post-message-form" onSubmit={onFormSubmit}>
+        <label htmlFor="newMessage"><h3 className="post-message-label">What's making you happy right now?</h3></label>
```
IsabellaM5, line 40: I would consider the usage of h3 here inside of a label tag. I don’t think this is a semantically correct way of using a h3. I would consider using a span tag if you want to wrap the text here in an element 🙂

code/src/components/PostMessage.js
```diff
@@ -0,0 +1,57 @@
...
+        <textarea className="post-message-textarea" 
+          id="newMessage"
+          placeholder="Write your happy thought.."
+          //type="text-area"
```
IsabellaM5, line 44: Opppss, maybe you should've cleaned this up? 😉

code/src/components/PostMessage.js
```diff
@@ -0,0 +1,57 @@
...
+          value={messageNew}
+          onChange={onMessageNewChange}
+          />
+        <div className="btn-container">
+          <button className="post-message-btn" type="submit"><span className="heart-emoji" role="img" aria-label="heart-emoji">❤️</span> Send happy thought <span className="heart-emoji" role="img" aria-label="heart-emoji">❤️</span></button>
+          <p>{chars}/150</p>
```
IsabellaM5, line 50: Nice solution for the counter function here! 💪

code/src/components/ListMessages.js
```diff
@@ -0,0 +1,40 @@
...
+      .then(messages => setMessageList(messages))
+      .catch(err => console.error(err))
+  }
+
+  useEffect(fetchMessageList, [])
+
+  return (
+    <>
+      {messageList.map(post => (
+        <article key={post._id} className="message-container">
```
IsabellaM5, line 21: Here I actually think an article tag is a viable option since you’re displaying the content as an independent message. 😀

IsabellaM5 (approved):
Fantastic job with this project! 😀 🐯

You got the basic functionality in place in accordance with the blue level requirements. You also implemented a counter below the input form that updates as the user types. Also, I loved the background gradient you did for this page ✨

Your project is hereby approved! 👍 For some specific comments/suggestions, see the comments made in the code files.

/Isabella

code/src/components/LikePost.js
```diff
@@ -0,0 +1,49 @@
...
+  const postHearts = (messageId) => {
+    fetch(LIKE_URL(messageId),{
+      method: 'POST',
+      headers: { "Content-Type": "application/json" }
+    }) 
+      .then (() => {
+      onLiked(messageId)
+      fetchMessageList()
+    }) 
+      .catch(err => console.error(err))
```
KarolCislowski, line 20: console outputs can't be present in the production code.

KarolCislowski (changes requested):
Hej!

KarolCislowski (approved):

PR #197, opened 2021-03-31
Week 11 Project

FD000:
Link to my deployed project:


code/src/App.js
```diff
@@ -1,9 +1,78 @@
...
+    fetch(API_URL)
+    .then(res => res.json())
+    .then(message => setMessageList(message))
+    .catch(err => console.error(err))
+  }
+
+  const OnInputMessage = (event) => { 
+    setMessageNew(event.target.value)
+  }
+
```
klimenko3007, line 28: A small comment here: 
the function names should use camel case and should not start with capital letter. So it should be onInputMessage

code/src/components/Form.js
```diff
@@ -0,0 +1,38 @@
...
+                    <textarea
+                    className={words === 140 ? "input red" : "input"}
+                    id="newMessage"
+                    type="text" 
+                    value={messageNew}
+                    onChange={OnInputMessage}
+                    cols="40" 
+                    rows="6"
+                    maxLength="140"
+                    placeholder="Please type your thoughts here. Maximum 140 characters allowed."
```
klimenko3007, line 23: I think this should be indented, otherwise it is hard to read.

code/src/components/Form.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+
+const Form = ({messageNew, OnInputMessage, onFormSubmit}) => {
+    let words = messageNew.length
```
klimenko3007, line 4: I think here you can use const because you never reassign the value to words

klimenko3007 (changes requested):
Congratulations, Fethullah, I think you did a really good job with this project!! 
I only noticed some minor things that need to be changed to make it even better!

code/src/App.js
```diff
@@ -1,9 +1,78 @@
...
+  useEffect(() => {
+    fecthMessageList()
+
+  }, [])
+
+  const fecthMessageList = () => {
+    fetch(API_URL)
+    .then(res => res.json())
+    .then(message => setMessageList(message))
+    .catch(err => console.error(err))
```
KarolCislowski, line 22: console outputs can't be present in the production code.

KarolCislowski (changes requested):
Hej!

KarolCislowski (approved):
😸

PR #221, opened 2021-10-29
anna-dimi-happy-thoughts

anndimi:
Link to my deployed project: https://happy-happy-thoughts.netlify.app


code/src/App.js
```diff
@@ -1,9 +1,63 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
+import { API_URL, LIKES_URL } from './utils/urls'
+import ThoughtForm from 'components/ThoughtForm'
+import ThoughtItem from 'components/ThoughtItem'
+import LoadingSpinner from 'components/LoadingSpinner'
+import ReloadButton from 'components/ReloadButton'
```
ebbadelsol, line 6: Nice structure for the components

code/src/components/ThoughtItem.css
```diff
@@ -0,0 +1,130 @@
...
+
+.heart-icon {
+    width: 85%;
+    margin-top: 5px;
+}
+
+.heart-icon:hover {
+    animation: pulse 1.5s infinite;
+    animation-timing-function: linear;
+}
```
ebbadelsol, line 61: Really nice animation!

If you want you could make the heart animation when you hover the parent like this. (It will still only make the animation for the heart but you can se the animation if you hover the parent) :

```
like-button:hover heart-icon {
    animation: pulse 1.5s infinite;
    animation-timing-function: linear;
}
```

ebbadelsol (approved):
You have achieved in making a very nice looking page with a lot of details that helps the user navigate thru the page at the same time as you are giving the user a pleasant experience.

I really like that you have disabled people from typing a thought longer than 140 characters (instead of just making the button disabled).

Really nice styling – as always!

The loading also looks very good for both the main page and for the form on submit.

PR #224, opened 2021-10-29
JohannaMJ Happy-thoughts-project

JohannaMJ:
The Happy Thoughts is a React-app where you can read, post and like happy thoughts. The list of thoughts comes from an API and displays the 20 latest thoughts that have been posted.

I started by thinking how to structure the app, then created components for the different parts of the project. Most functions are situated in the parent component, and relevant information is sent to the child components via props. I used state to store and update objects properties, as well as useEffect to fetch data from an API whenever the component is mounted. In this way, new posts are displayed when submitting the form, and the amount of likes increase when clicking the heart-button.

If I had more time I would create an animation for when a new thought is submitted and appears in the list below

https://share-some-happiness.netlify.app/


jakobxlindstrom (commented):
Great work Johanna! 
You have made a great React app! 
Fulfilling more than blue requirements and showing great accessibility skills! 

Your design is as is should be and your code is easy and clear to follow! 

Keep up the great work!

PR #234, opened 2021-10-30
Happy Thoughts Rebecca Blixt

rebeccablixt:
Link to my deployed project:
https://britishswede-happy-thoughts.netlify.app/

I felt a little bit unmotivated this week due to non-coding related things. I am happy that I managed to create the project and that it follows the given design. Also happy that I was able to add some extra features, like the loading screen. 

Nothing really, I think I've got the hang of it!

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,65 @@
-import React from 'react'
+import React, { useEffect, useState } from "react";
+import moment from "moment";
+
+import { NewPost } from "./components/NewPost";
+// import Posts from "./components/Posts";
+import { API_URL } from "./utils/urls";
+import { LIKES_URL } from "./utils/urls";
 
 export const App = () => {
+  const [thoughts, setThoughts] = useState([]);
+  const [newThought, setNewThought] = useState("");
+
+  useEffect(() => {
+    fetch(API_URL)
+      .then((res) => res.json())
+      .then((data) => setThoughts(data));
+  }, []);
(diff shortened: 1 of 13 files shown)


code/src/index.css
```diff
@@ -11,3 +12,147 @@ code {
...
+  font-size: 14px;
+}
+
+.form {
+  background-color: #f2f0f0;
+  position: relative;
+}
+
+/* Input */
+#newThought {
```
nama0027, line 85: even though using Ids is not wrong at all however it is consider coding convention that classes should be used for styling and Ids should left for Javascript

code/src/App.js
```diff
@@ -1,9 +1,75 @@
...
+    };
+
+    fetch(API_URL, options)
+      .then((res) => res.json())
+      .then((data) => {
+        fetchThoughts();
+      });
+  };
+
+  const onLikesIncrease = (thoughtId) => {
```
nama0027, line 44: as a reminder of coding convention again :) if the function is going to be passed as props than its a convention to start it with 'handle' instead of 'on'

nama0027 (commented):
job well done! 

The happy thoughts page is working so good. you have fulfilled all blue requirements and the red requirements. well done!

The code is well organised and so clear to follow. its good that you have created separate component for the button also. its a good approach to follow in my opinion.
keep the good work coming!
/Naushin

PR #235, opened 2021-10-30
Week 11 - Project-happy-thoughts

Fatima-GR:
# Technigo React Starter App

During this project, I got introduced to the full concept of the component lifecycle, how to use useEffect hook in React to make components mount, unmount, or when state changes. This project has been very helpful for me to understand more about React components, usage of states, and sending data through functions and props.

## Link to my deployed project: 
https://nifty-blackwell-f65e37.netlify.app/


code/public/index.html
```diff
@@ -1,9 +1,12 @@
 <!DOCTYPE html>
 <html lang="en">
-
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link
```
Meeteyes, line 6: Nice, now I know where to import the fonts in React :)

code/src/components/Main.js
```diff
@@ -0,0 +1,76 @@
+import React, { useEffect, useState } from "react";
+import ThoughtForm from "./ThoughtForm";
+import ThoughtCard from "./ThoughtCard";
+import Spinner from "./Spinner";
+import { API_URL, LIKES_URL } from "../utils/urls";
```
Meeteyes, line 5: Great to see that you are using the utils approach.

code/src/components/Main.js
```diff
@@ -0,0 +1,76 @@
...
+
+const Main = () => {
+  /* useStates */
+  const [thoughts, setThoughts] = useState([]); //It b
+  const [newThought, setNewThought] = useState(""); // It being used in the form to send the new message
+  const [loading, setLoading] = useState(false);
+
+  /* useEffects: it renders after the component gets mounted */
```
Meeteyes, line 13: Good that you wrote a fetch function. It keeps the code very neat and clear.

code/src/components/Main.js
```diff
@@ -0,0 +1,76 @@
...
+  useEffect(() => {
+    fetchThoughts();
+  }, []);
+
+  const fetchThoughts = () => {
+    setLoading(true);
+    fetch(API_URL)
+      .then((res) => res.json())
+      .then((data) => setThoughts(data))
+      .finally(() => setLoading(false));
```
Meeteyes, line 23: I haven't worked with .finally so far. What are the advantages of using finally instead of setting the loading to false in the 2nd .then?

code/src/components/ThoughtCard.js
```diff
@@ -0,0 +1,33 @@
...
+        <p>{thought.message}</p>
+      </div>
+
+      <div className="heart-date-container">
+        <button
+          className="heart-btn"
+          onClick={() => onLikesIncrease(thought._id)}
+          //the heart-btn background-color depends on the likes or not
+          style={{
+            backgroundColor: thought.hearts > 0 ? "#FFADAD" : "#EAEAEA",
```
Meeteyes, line 17: This is a nice styling detail!

Meeteyes (commented):
Hi, Fatima!
Very nice work. The code is super well organized and understandable. I didn't notice any repetitious or redundant lines. Very good! I also was intrigued about the .finally method in the fetch request which I haven't used so far.

Best,
Reinis

code/src/components/Main.js
Fatima-GR: Good question Reinis!  As far as I understand,  the methods then and finally are almost the same with the exception that when there is an error, the method finally will always execute but "then" will not. In this particular case, I am not sure either the advantage of setting the SetLoading in the finally function, I followed Maxsymilian's example but I am so intrigued now so I will ask this question in the FQ session on Friday so hopefully, we get the answer.

Fatima-GR (commented):
Thank you for doing my code review Reinis! I hope we get a clear answer on Friday about the advantage of using the .finally() function.

PR #268, opened 2021-10-31
Happy-thoughts-Eddie

Lundgreneddie:
Link to my deployed project:
https://happy-thoughts-eddie.netlify.app/


code/src/index.css
```diff
@@ -1,13 +1,89 @@
 body {
```
Mattsson717, line 1: A tip for keeping track of all the colors, since its basicly the same colors multiple times, is to use this on top of the css:

:root {
  --primary: #181818;
  --secondary: #ffffff;
  --tertiary: #a0a0a0;
  --tertiaryopacity: #a0a0a083;
}

Then you can use the colors like this: color: var(--primary)
so you only need to change colors in one place :)

Mattsson717 (commented):
Nice work on the project. Its easy to follow and the structure and semantics of the code looks great.

Be proud!

PR #270, opened 2021-11-01
Nina Rivera - Penny for your thoughts

NinaAlejandra:
Link to my deployed project:
https://ninarivera-pennyforyourthoughts.netlify.app/
It was very difficult, and I'm not happy with how it looks, but I didn't have time for more. 
It could and should look better than this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 # Happy Thoughts
 
-Replace this readme with your own information about your project.
+Our very own Twitter, but kinder.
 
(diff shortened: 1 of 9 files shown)


hejmaria (commented):
Looks great Nina! You should be proud of this project, you did a really nice job! Everything works and you have met the requirements. If you want, you could add just a few media queries because the app looks a bit squished in mobile-view, and in iPad pro (width 1024px) the containers run outside of the screen. Looks great on desktop and iPad.

PR #275, opened 2021-11-03
Emelie Söderhielm Happy Thoughts

intehon:
# Happy Thoughts

Project to practice using effect and state in React

## View it live

https://intehon-happy-thoughts.netlify.app/


code/src/App.js
```diff
@@ -1,9 +1,97 @@
...
+    event.preventDefault()
+
+    const options = {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({ message: newThought }),
+    }
+    event.target.reset()
```
sofiawillebrand, line 47: Nice, I didn't know that you could use something called reset!

code/src/App.js
```diff
@@ -1,9 +1,97 @@
...
+
+    fetch(API_URL, options)
+    .then((res) => res.json())
+    .then((data) => {
+    fetchThoughts()
+    })
+
+    // Clears input-field and let's you post a new thought
+    setNewThought('')
```
sofiawillebrand, line 56: Hm, I might be misunderstanding the code here, does this do the same as the function on 47? In that case, one of them can be removed.

code/src/index.css
```diff
@@ -1,13 +1,195 @@
...
+}
+
+.heart-btn-clicked {
+  background-color: rgb(255, 173, 173);
+}
+
+.heart-btn:hover {
+  width: 43px;
+  height: 43px;
+}
```
sofiawillebrand, line 132: To avoid making the text-box getting higher when hovering on the button, you could have removed as much margin from the button as the button grows.

sofiawillebrand (commented):
The application: I love the styling, it looks like the styling we were given but with some nicely added details. Impressive with the counter of how many times one have liked a post! The loader was a neat function, I like how the page is blank except for the loader. 

The text-boxes weren't really responsive for smaller screens, they disappear outside of the page.

The code: The use of comments is a plus, it makes it easier to follow what is happening. It was difficult to find something to comment on, which I think is a good sign! 

Nice work with the assignment :D

code/src/App.js
intehon: Haha, you might be right 😅 will check it out!

PR #277, opened 2021-11-03
Happy Thoughts - Lena S

lenisili:
Link to my deployed project:

### DIFF
--- a/code/src/index.css
+++ b/code/src/index.css
@@ -5,9 +5,187 @@ body {
     sans-serif;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  background-size: cover;
 }
 
 code {
   font-family: source-code-pro, Menlo, Monaco, Consolas, "Courier New",
     monospace;
+  font-family: 'Roboto Mono', monospace;
+}
+
+.body{
+  width: 100%; 
+}
+
+.date {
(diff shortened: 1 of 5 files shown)


Alisebrink (commented):
Hi Lena! It looks like you have made a good project :)

I have a few pointers though. I can see that you only have made two commits and I would really recommend you to do a few more, try to commit all the time in case your computer crashes and everything goes away. Committing often also helps with error in case you write something that breaks your website you can always go back to the previous commit and see what you changed. 

I also would recommend you to write more comments in your code. Not only for yourself but also for the one that are reviewing your code. It would make it easier to follow along and if you work at a company sometimes your colleges take over your code, then it is also very nice to have comments. It is a good practice to write many comments :)

And also when you do your pull request include a link to your deployed website on Netlify to make it easier to view, that also makes it easier to comment on your code if you can see it in practice :)

But all in all, great project!

code/src/index.css
```diff
@@ -5,9 +5,187 @@ body {
...
+
+.heart-btn:hover {
+	transform: scale(1.1);
+}
+
+.date {
+	color: #949494;
+	display: flex;
+	align-items: center;
+}
```
Alisebrink, line 191: Only comment I have is that you didnt code with a mobile first approach which is a good practice :) media queries are missing in this project

PR #291, opened 2022-03-25
Jin - w11

jiiinCho:
Link to my deployed project:
https://technigo-happythoughts-2022.netlify.app/

There is a delay issue to update when user click 'like' button, because it is asynchronous request. This has been solved by using setTimeout Web API.

CSS white-space: wrap does not work if there is no space in content. That caused display issues when message is overflowing. Current solution set 'hidden' if the content overflow, but this might not be the best way to display message.


code/src/App.js
```diff
@@ -1,9 +1,63 @@
...
+    const timeout = setTimeout(() => {
+      httpRequest("", { method: "GET" })
+        .then((data) => {
+          console.log("fetch");
+          const sorted = data.sort((a, b) => {
+            const curr = new Date(a.createdAt).getTime();
+            const next = new Date(b.createdAt).getTime();
+            return next - curr;
+          });
+          setList(sorted);
```
idanaslund, line 36: Great that you sorted the sent messages here! As you said before it's good to practice!

code/src/components/Form.js
```diff
@@ -0,0 +1,57 @@
...
+
+const Form = ({ handleFormSubmit }) => {
+  const [message, setMessage] = useState("");
+  const [alert, setAlert] = useState(false);
+  const onMessageChange = (e) => {
+    const count = message.length;
+    if (count > 140) {
+      setAlert(true);
+    } else {
+      setAlert(false);
```
idanaslund, line 11: I thought this was going to display an alert if I tried to send a message with more than 140 characters. But is this only for the message not to send when it is more than 140?

code/src/components/Form.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+    setMessage(e.target.value);
+  };
+
+  const onFormSubmit = (event) => {
+    event.preventDefault();
+    if (message.length < 140) {
+      handleFormSubmit(message);
+      setMessage("");
```
idanaslund, line 20: I like that it isn't possible to send when you go above 140 characters in the message, also the text becomes red. Very clear for the user.

code/src/components/Form.js
```diff
@@ -0,0 +1,57 @@
...
+  const [message, setMessage] = useState("");
+  const [alert, setAlert] = useState(false);
+  const onMessageChange = (e) => {
+    const count = message.length;
+    if (count > 140) {
+      setAlert(true);
+    } else {
+      setAlert(false);
+    }
+    setMessage(e.target.value);
```
idanaslund, line 13: Maybe you are already aware or maybe I misunderstand the purpose, but there is no message alerted when I reach 140 or try to send more than 140 characters.

idanaslund (approved):
I really like your project! It looks to the point the same as the design and you have achieved red level and touched some of the black levels. 

One thing I noticed was that no alert was displayed when trying to send a message with more than 140 characters, maybe it's something you are working on?

I feel like I really learned alot from your code, thank you for that! Great work with this project, you should be proud!

PR #293, opened 2022-03-25
Happy thoughts <3

rawisou:
# Happy Thoughts

## View it live

https://illustrious-frangipane-e38a00.netlify.app/


code/src/index.css
```diff
@@ -1,13 +1,129 @@
...
+  display: inline-block;
+  border: none;
+  -webkit-appearance: none;
+  margin: 10px 0 10px 20px;
+  padding: 5px 15px 5px 15px;
+  border-radius: 20px;
+}
+
+.send-btn:hover {
+  border: 2px solid pink;
```
EmmaSprings, line 70: Perhaps this border could be removed, or add a border with the same appearance, but grey, to .send-btn so that when you hover, it wont "jump".

EmmaSprings (commented):
Great work on the happy thought project! Your code is easy to follow and the site works as it is supposed to. 
I left a comment on the send button because I noticed the size increases when hovering over it which makes it "jump". Other than that all is very well done. I like the count of characters when writing aswell, very nice! Definitely meets requirements!

code/src/index.css
rawisou: Thank you so much!! I removed it and now it stopped jumping. 😀

PR #294, opened 2022-03-25
Week 11 - Michael

michaelchangdk:
Link to my deployed project:


NabeelMansour (commented):
Nice done as usual, there is though some styles in the css file that are declared but not used. Well done Michael

PR #306, opened 2022-03-25
Project happy thoughts week.11

Mimmisen:
Link to my deployed project: https://happy-thoughts-mimmi.netlify.app/


qwer-ctrl (commented):
Nice job Mimmi! I like your animation fot sending button, it is lovely :) The code is simple and easy to read, you have reach all blue level requirements and also things from higher levels. You have spend time for those animations and they make the site more comfortable to use. I didn't notice any bugs while trying the app.

PR #310, opened 2022-03-26
Happy Thoughts - Arianna

ariallahyar:
Link to my deployed project: https://feelgoodfeed.netlify.app/


Nosslexa (commented):
Great job with this weeks project! Im so impressed with how you manage to get so much done with so little code. Everything works as it should and you managed to get the like buttons to change color after getting clicked. Also great job adding styled components.

PR #311, opened 2022-03-26
Happy Thoughts by Hanna

HWallberg:
Link to my deployed project: https://cute-hummingbird-6688bb.netlify.app/

It's so cool that I made it work I must say! I'm really happy with the result. If I had more time I would have fixed so the input field cleared on button submit. 

I had big issues with pushing this to github so spent quite some time on that. Solved it by download GitHub Desktop. It's like the terminal but with folders instead.

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,21 @@
-import React from 'react'
+import React, { useState } from 'react'
+
+import Header from 'components/Header'
+import NewThought from 'components/NewThought'
+import ThoughtList from 'components/ToughtList'
 
 export const App = () => {
+  const [thoughtList, setThoughtList] = useState ([])
+  const [newThought, setNewThought] = useState ('')
+
   return (
+    <>
     <div>
-      Find me in src/app.js!
+      <Header />
+      <NewThought newThought={newThought} setNewThought={setNewThought} />
+      <ThoughtList thoughtList={thoughtList} setThoughtList={setThoughtList} />
     </div>
+    </>
+    
   )
 }
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -5,12 +5,12 @@ import NewThought from 'components/NewThought'
 import ThoughtList from 'components/ToughtList'
 
 export const App = () => {
-  const [thoughtList, setThoughtList] = useState ([])
   const [newThought, setNewThought] = useState ('')
+  const [thoughtList, setThoughtList] = useState ([])
 
   return (
     <>
-    <div>
+    <div className="main-div">
       <Header />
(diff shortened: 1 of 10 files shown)


code/src/App.js
```diff
@@ -1,9 +1,21 @@
-import React from 'react'
+import React, { useState } from 'react'
+
+import Header from 'components/Header'
```
vanhaj, line 3: Great job doing that extra job of creating a header component!

code/src/components/NewThought.js
```diff
@@ -0,0 +1,53 @@
...
+            .then((data) => setThoughts([data, thoughts])) 
+    }
+
+    return (
+        <>
+            <form className="form" onSubmit={onFormSubmit}>
+                <label>What's making you happy right now?</label>
+                <textarea 
+                    className="input-field"
+                    rows="4"
```
vanhaj, line 37: This is a great way of making the input field bigger! I forgot to add that one myself

code/src/index.css
```diff
@@ -5,9 +5,103 @@ body {
...
+h1 {
+  text-align: center;
+  margin: 5%;
+}
+
+.form {
+  display: flex;
+  flex-wrap: wrap;
+  min-width: 320px;
+  max-width: 400px;
```
vanhaj, line 26: What a clever way of solving the width for the boxes, taught me something new!

code/src/index.css
```diff
@@ -5,9 +5,103 @@ body {
...
+    box-sizing: border-box;
+    margin-bottom: 10px;
+    width: 100%;
+    padding: 2%;
+}
+
+label,
+p,
+h1,
+h3 {
```
vanhaj, line 94: I think that this is such an awsome way of reusing them for the same purpose!

code/src/index.css
```diff
@@ -5,9 +5,103 @@ body {
...
+
+.post-details {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  color: #b0b0b0;
+  margin-top: 5%;
+}
+
+.like-button {
```
vanhaj, line 63: Just wanted to add a suggestion. You could change the cursor to pointer instead like this:
`cursor: pointer;`
Or add a hover effect to the button so that you can see that the button is clickable. You could also add it to the .submit-button :)

vanhaj (approved):
I think your code is really clean, and I find it very easy to follow. 
You already mentioned this, but it whould have been a nice add on if you managed to clear the input field, but don't worry I dont find it a big issue! Everything else is working really well and I think you managed to get the design really look alike as the pre made one. 
You did a really great job Hanna! Thank you for sharing you code, I've learned a lot just by looking at it!

PR #313, opened 2022-03-26
Uploading the project

NabeelMansour:
Link to my deployed project:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,11 @@
 # Happy Thoughts
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+The assigment is about fetching and using the happy-thoughts api, with the help of React applying it to the UI.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+The approach was by React, momment.js for the time upload and some tachyons for some style. If I had more time, the next would be some animations, but it was a bit difficult.
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+https://grand-paletas-f56f32.netlify.app/
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,17 @@
 import React from 'react'
+import { TaskForm } from './Components/TaskForm'
+
+
 
 export const App = () => {
+  
   return (
+
     <div>
-      Find me in src/app.js!
-    </div>
+      <TaskForm />
+     </div>
+
+     
+
   )
 }
--- a/code/src/Components/TaskForm.js
+++ b/code/src/Components/TaskForm.js
@@ -20,7 +20,7 @@ export const TaskForm = () => {
 
   const getThoughts = () => {
       // setLoading(true);
-      fetch('https://happy-thoughts-technigo.herokuapp.com/thoughts')
+      fetch('https://projecthappythoughtsapi.herokuapp.com/thoughts')
       .then(res => res.json())
       .then(data => setThoughts(data))
       // .finally(() => setLoading(false))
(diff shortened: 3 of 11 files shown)


code/src/App.js
```diff
@@ -1,9 +1,17 @@
 import React from 'react'
+import { TaskForm } from './Components/TaskForm'
```
amandatange, line 2: This could maybe be called something that is more sematically cohesive with what it contains.

code/src/Components/TaskForm.js
```diff
@@ -0,0 +1,84 @@
...
+        },
+        body: JSON.stringify({
+            message: newMessages
+        })
+      }
+
+    fetch('https://happy-thoughts-technigo.herokuapp.com/thoughts', options)
+    .then(res => res.json())
+    .then(() => getThoughts())
+    .finally(()=> setNewMessages(''))
```
amandatange, line 46: The submitting of new messages works well but maybe consider also setting an interval for fetching new thoughts every 5 or so seconds to retrieve thoughts that others might post.

code/src/Components/TaskForm.js
```diff
@@ -0,0 +1,84 @@
...
+    <div>
+            <ThoughtForm
+              onFormSubmit={handleOnFormSubmit}
+              newMessages={newMessages}
+              onSetThoughtChange={handleOnNewThought}
+                
+            />
+    
+  
+    {thought.map((thoughts) => (
```
amandatange, line 74: I would consider switching names for the array of thoughts to 'thoughts' and the singular thought referred to as thought, but that only to make it a little easier to understand semantically.

code/src/Components/ThoughtElements.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import Moment from 'moment';
+
+
+
+const ThoughtElement = ({ thought, handleLike }) => {
+  console.log(thought)
```
amandatange, line 7: There's only one console log here but all 20 thoughts are logged when there is a new post so it quite quickly becomes a very full console.

code/src/Components/ThoughtForm.js
```diff
@@ -0,0 +1,28 @@
...
+    
+    <form className="grow shadow-20" onSubmit={onFormSubmit}>
+      <h3>Any good thoughts to share today ?</h3>
+
+      <textarea 
+      placeholder='Any thoughts...' 
+      rows="3" cols="30" 
+      value={newMessages} 
+      onChange={onSetThoughtChange}
+      />
```
amandatange, line 15: when testing for accessibility the Wave report suggests using a label element or an aria description for the textarea element

code/src/Components/ThoughtElements.js
```diff
@@ -0,0 +1,31 @@
...
+
+
+  return (
+    <section className='thought-container grow shadow-20'>
```
amandatange, line 11: don't know really how to comment on the hover effect for the thoughts containers since it's not done in the css file but i think maybe that the scaling makes so that the like button is only clickable on the right side of the button. When the left side of the like button is pressed the container shinks and then grows again which is also quite abrupt.

code/src/Components/ThoughtElements.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import Moment from 'moment';
```
amandatange, line 2: Consider using another library, like date-fns, since Moment was deprecated and will eventually become obsolete.

README.md
```diff
@@ -1,13 +1,11 @@
...
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+The assigment is about fetching and using the happy-thoughts api, with the help of React applying it to the UI.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+The approach was by React, momment.js for the time upload and some tachyons for some style. If I had more time, the next would be some animations, but it was a bit difficult.
```
amandatange, line 7: would be nice to see some example of a problem and a solution that you encountered and perhaps a little more description of the project that doesn't require you to know what the happy thoughts api is.

amandatange (commented):
Over all very good project, the site looks good and works for the most part, except for that like button issue with the left side of the button becoming unclickable. The code is easy to follow and well structured but since the styling was done not only in css it was not possible to feedback on that. Well done!

PR #317, opened 2022-03-26
Project w.11 - Happy thoughts - Jennifer Wang Kurtto

jenniferwvng:
Link to my deployed project:
https://aquamarine-praline-88ba8a.netlify.app/

Good, more to learn about useEffect though. Could be optimized when I have made research on how to do it.

UseEffect, async functions, displaying error handlers, how to not have to fetch several times due to network costs, but still retrieve other peoples messages from the API


code/src/components/Form.js
```diff
@@ -0,0 +1,76 @@
...
+          <h1 style={{ color: 'pink' }}>What&apos;s making you happy right now?</h1>
+          <p style={{fontSize: '12px', wordWrap: 'break-word'}}>{errorMessage}</p>
+          <input
+            id="form"
+            type="text"
+            name="message"
+            value={userInput}
+            onChange={updateUserInput} />
+            <p>
+              <span style={{color: userInput.length >= 140 ? 'red' : 'green' }}>
```
rawisou, line 58: Nice use of the conditional rendering

code/src/index.css
```diff
@@ -11,3 +11,96 @@ code {
...
+
+  .form-button {
+    font-size: 15px;
+    margin-top: 10px;
+  }
+}
+
+@media screen and (min-width: 992px) {
+  .form-div {
+    margin: 30px;
```
rawisou, line 89: Very small thing, line 89 and 90 are the same as line 71 and 72.

code/src/index.css
```diff
@@ -11,3 +11,96 @@ code {
...
+    width: 30vw;
+    height: 60vh;
+  }
+}
+
+@media screen and (min-width: 1200px) {
+  .form-div {
+    margin: 30px;
```
rawisou, line 97: Same here with line 97 and 98, same as line 71 and 72. I'm not sure if we remove it, things will stay the same (I think it should).

rawisou (commented):
Your site works as it should. 

The use of async and await in your function is nice. 
Very small duplicates in index.css but this does not affect how your site works. 

Great job!

PR #320, opened 2022-03-27
Happy Thoughts - Emma Springs

EmmaSprings:
Link to my deployed project:
https://musical-sopapillas-20a48e.netlify.app/?

Please see the README file


code/src/App.js
```diff
@@ -1,9 +1,13 @@
...
+import Collection from 'components/Collection'
+
+
 export const App = () => {
+
   return (
     <div>
-      Find me in src/app.js!
+      <Collection />
     </div>
```
ariallahyar, line 11: Because there is nothing else happening in `App`, I feel like you could almost move everything from `Collection` into app to avoid somewhat redundant files. Not sure though, I'm no expert in structuring React.

code/src/components/Collection.js
```diff
@@ -0,0 +1,89 @@
...
+
+import API_URL from 'utils/Api'
+
+import Message from 'components/Message'
+import SentMessage from 'components/SentMessage'
+
+
+const Collection = () => {
+
+    const [fetchThought, setFetchThought] = useState([])
```
ariallahyar, line 11: Naming convention could be a bit more descriptive to what this is actually doing and makes sense in all places it's being used. So instead of `fetchThought`, something like this:
```javascript
const [messages, setMessages] = useState([]);

File: code/src/components/Collection.js [hunk 3]
```diff
@@ -0,0 +1,89 @@
...
+
+    const handleOnLikeChange = (likeID) => {
+
+        const options = {
+            method: 'POST',
+        }
+
+        fetch(`https://happy-thoughts-technigo.herokuapp.com/thoughts/${likeID}/like`, options)
+            .then(res => res.json())
+            .then(() => fetchThoughts())
```
ariallahyar, line 65: I don't think `fetchThoughts` is doing anything here.

code/src/components/Collection.js
```diff
@@ -0,0 +1,89 @@
...
+    return (
+        <div>
+            <Message
+                loading={loading}
+                onNewMessageChange={onNewMessageChange}
+                onFormSubmit={onFormSubmit}
+            />
+
+            <SentMessage
+
```
ariallahyar, line 79: Small things, there are some extra spaces here and there. **Prettier** is a helpful extension to look into. I think VS Code's built-in formatter might also do the trick. Try **option + shift + F** :)

code/src/index.css
```diff
@@ -1,13 +1,150 @@
...
-  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen",
-    "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue",
-    sans-serif;
+  box-sizing: border-box;
+  margin: 25px 0 0 0;
+  font-family: 'Inconsolata', monospace;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+
+  /* animated background */
```
ariallahyar, line 8: Love the animated gradient! I had no idea how to do this.

code/src/components/Message.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+
+const Message = ({ loading, newMessage, onNewMessageChange, onFormSubmit }) => {
+
+    // SHOWN WHEN IMAGE RELOADS WHICH I HAVENT MANAGED TO FIX, DO NOT WANT IT TO SHOW..
+
+    if (loading) {
+        return <div className='loading-container'>
+            <img className='loading' src='./icons/heart.png' alt="heart icon" />
+        </div>
```
ariallahyar, line 10: Not sure I completely follow, but a little insight: the way your code is structured, I believe the entire page renders again every time a message is submitted. Assuming that's correct, you don't currently have a way of avoiding showing the loading heart when a message is submitted. It will happen both when you first load the page and when you submit because essentially the page starts all over again by reloading everything.

ariallahyar (commented):
Looks great, Emma! Your project was easy to follow, and meets the requirements. Nice job!

PR #323, opened 2022-03-27
Dorothea's happy thoughts

Dorothea-B:
Link to my deployed project:

happy-thoughts-thea.netlify.app

I'm satisfied w. my project, but didn't have the energy to move my code into individual components or finish the black level this week. But I'm getting more comfortable with react and experimenting and trying things out on my own!  

Oftentimes when I'm googling for answers or browsing stack overflow, the solutions are using older syntax. It would be nice to get an explanation of react-classes, super(), "this.state" and stuff like that, so I can better understand and translate this into the newer syntax that we're using.

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -25,7 +25,7 @@ export const App = () => {
     
     fetch( API_URL, 
         { method: 'POST', headers: { 'Content-Type': 'application/json',}, body: JSON.stringify({ message: newThought }) }
-      ).then((res) => res.json()).then((data) => setThoughts([data, ...thoughts])).then(setNewThought(event.target.value))
+      ).then((res) => res.json()).then((data) => setThoughts([data, ...thoughts])).then(setNewThought(newThought))
 
       if (newThought.length < 4 ) {
         alert('Too few characters')
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -36,7 +36,6 @@ export const App = () => {
       } 
     }
 
-    useEffect(() => { setThoughts(thoughts) }, [onFormSubmit])
 
 
   const addLike = (thoughtId) => {
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -25,7 +25,7 @@ export const App = () => {
     
     fetch( API_URL, 
         { method: 'POST', headers: { 'Content-Type': 'application/json',}, body: JSON.stringify({ message: newThought }) }
-      ).then((res) => res.json()).then((data) => setThoughts([data, ...thoughts])).then(setNewThought(''))
+      ).then((res) => res.json()).then((data) => setThoughts([data, ...thoughts])).then(setNewThought(event.target.value))
 
       if (newThought.length < 4 ) {
         alert('Too few characters')
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -34,7 +34,6 @@ export const App = () => {
       else if ( newThought.length > 140 ) {
         alert('Too many characters')
       } 
-      setThoughts(thoughts)
     }
 
     useEffect(() => { setThoughts(thoughts) }, [onFormSubmit])
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -20,20 +20,24 @@ export const App = () => {
   }, [] )
 
 
+
   const onFormSubmit = (e) => {
-    e.preventDefault()
+    // e.preventDefault()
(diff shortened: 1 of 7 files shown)


Dorothea-B:
I updated my code a little after doing my pull req, don't know if there's a way to update my commits within the pull?

code/src/App.js
```diff
@@ -1,9 +1,115 @@
...
+    .then(data => setThoughts(data))
+
+  }, [] )
+
+
+  const onFormSubmit = (e) => {
+    e.preventDefault()
+    
+    fetch( API_URL, 
+        { method: 'POST', headers: { 'Content-Type': 'application/json',}, body: JSON.stringify({ message: newThought }) }
```
sukiphan97, line 27: I would recommend creating a variable name options to save this line of code. It will be easier for you to keep track and keep the code clean

code/src/App.js
```diff
@@ -1,9 +1,115 @@
...
+      ).then((res) => res.json()).then((data) => setThoughts([data, ...thoughts])).then(setNewThought(''))
```
sukiphan97, line 28: We dont need to write 2x 'then' in this case. You can combine them together like this: 
`then((data) => { 
setThoughts([data, ...thoughts];
setNewThought('')
})`

sukiphan97, line 28: To get the error message, in case the fetch is not successful. You can add `catch(error => console.log(error) `right after `.then`

code/src/App.js
```diff
@@ -1,9 +1,115 @@
...
+    <div className="thoughts-wrapper">
+
+    <form onSubmit={onFormSubmit}> 
+      <label htmlFor="newThought">What's making you happy right now?</label>
+      <input 
+      className={characterStyling(characters)} 
+      value={newThought} 
+      onInput={(e) => setCharacters(e.target.value)}
+      onChange={(e) => setNewThought(e.target.value)}
+      id={newThought}
```
sukiphan97, line 77: To limit the input value, we can use html minlength and maxlength attributes. Check out the link below for more info: [https://www.w3schools.com/tags/att_input_maxlength.asp](url)

code/src/App.js
```diff
@@ -1,9 +1,115 @@
...
+      />
+     <p className="date"> Characters left: {140 - characters.length}</p>
```
sukiphan97, line 79: Nice way to add the counter

code/src/App.js
```diff
@@ -1,9 +1,115 @@
...
+        { method: 'POST'}
+
+    ).then((res) => res.json()).then((data) => {
+
+      const updateLikes = thoughts.map((item) => {
+          if(item._id === data._id) {
+          item.hearts += 1
+          return item }
+          
+          else { return item }
```
sukiphan97, line 51: you can remove else, and write only return item. It's unnecessary

code/src/App.js
```diff
@@ -1,9 +1,115 @@
...
+      <input 
+      className={characterStyling(characters)} 
+      value={newThought} 
+      onInput={(e) => setCharacters(e.target.value)}
+      onChange={(e) => setNewThought(e.target.value)}
+      id={newThought}
+      />
+     <p className="date"> Characters left: {140 - characters.length}</p>
+
+      <button  className="submit-btn" type="submit"><span role="img" aria-label="heart">💜</span> Send Happy Thought <span role="img" aria-label="heart">💜</span></button>
```
sukiphan97, line 81: Really nice! I like how you add aria-label here to improve accessibility

code/src/App.js
```diff
@@ -1,9 +1,115 @@
...
+        <div>
+          <button className={ thought.hearts === 0 ? 'notLikedClass' : 'likedColorClass' }
+          onClick={() => addLike(thought._id)}>
+          <span role="img" aria-label="heart">💜</span>
+          </button>
+        </div> 
+
+        <div>
+          x {thought.hearts}
+        </div>
```
sukiphan97, line 104: I am not sure why we need a dive to wrap the text here. Another way is to place the code line 103 after the <button> (line 99).

sukiphan97 (commented):
Hi Dorothea,

Sorry for the late review. Your page looks really nice, and the code is clean and easy to follow. Also, you added the media query which is a big plus for the project. I have few comments below, hopefully it will be helpful 😊

Suki

PR #344, opened 2022-10-06
Week7-Project-Happy-Thought.Sofie-Rydmark

SofieRydmark:
Link to my deployed project: https://w7-happy-thoughts.netlify.app/

My reflections on how this week's project turned out: Learned a lot more in depth of how react components actually work


code/src/App.js
```diff
@@ -1,9 +1,12 @@
 import React from 'react';
+import ThoughtWrapper from 'components/ThoughtWrapper';
+import Header from './components/Header';
```
carimore, line 3: This was a great idea to keep the App.js clean just to import the two components :)

code/src/components/NewThoughts.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+const NewThoughts = ({ newThought, handleSubmit, newThoughtChange, input }) => {
+  return (
+    <div className='newthought-container'>
+      <h2 className='newthought-header'>What`s making you happy right now?</h2>
+      <form onSubmit={handleSubmit}>
```
carimore, line 7: Using the Eslint in this case would work beacause it gives indentations for each object that is inside another object.

code/src/components/ThoughtWrapper.js
```diff
@@ -0,0 +1,83 @@
...
+      .finally(() => fetchThought());
+  };
+
+  return (
+    <>
+      <NewThoughts
+        newThought={newThought}
+        handleSubmit={handleSubmit}
+        newThoughtChange={newThoughtChange}
+        input={input}
```
carimore, line 72: Is this necessary as a separate from new thought because of the number of characters? Could they be combined?

carimore (commented):
It is really hard to review your work because it is so good, it is organized, clear and you always come up with creative solutions to problems. I just encourage you to try the eslint next time ;)

PR #349, opened 2022-10-07
Project Happy Thoughts, Cecilia F,

Sneezan:
✨**REFLECTIONS** ✨ 
This week has been hard for me in my personal life, which has led to this project not being as elaborate as I first initialised.  

Link to my deployed project:
https://onlyhappythoughts.netlify.app

The project turned out OK, it meets the blue level requirements as well as some red level. I am happy that I manage to deliver this week, but I am not happy about the project itself. I didn't follow the color scheme but tried to follow the design but in darker colors, to reflect my current state. 

I found the code this week challenging, mostly the POST and the .map arrays, so I spent a lot of time on that. 
If I had more energy this week I would like to add a loading function that doesn't render all the messages directly.  I had issues with load function, because the whole page glitched when updated then. If any has insights or solved that, please share. For now it's omitted. 

Screenshot 2022-10-07 at 14 45 39

### DIFF
--- a/code/src/components/Feed.js
+++ b/code/src/components/Feed.js
@@ -1,4 +1,3 @@
-/* eslint-disable */ 
 import React, { useEffect, useState } from 'react'
 
 import GeneratedFeed from 'components/GeneratedFeed.js';
@@ -31,7 +30,7 @@ const Feed = () => {
     fetchThoughts()
   }, [])
 
-  // POST FOR NEW THOUGHT 
+  // POST FOR NEW THOUGHT
   const handleFormSubmit = (event) => {
     event.preventDefault()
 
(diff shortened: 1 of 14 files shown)


code/src/components/PostNewThought.js
```diff
@@ -0,0 +1,47 @@
...
+      </section>
+      <textarea
+        rows="4"
+        id="newThought"
+        type="text"
+        placeholder="* good vibes only*"
+        value={newThought}
+        onChange={(e) => setNewThought(e.target.value)} />
+
+      <p className="letter-counter">{140 - newThought.length} / 140</p>
```
petrasson, line 32: Nice one!

code/src/components/PostNewThought.js
```diff
@@ -0,0 +1,47 @@
...
+      <div className="buttons">
+        <button
+          className="send-button"
+          disabled={newThought.length < 1 || newThought.length > 140}
```
petrasson, line 36: I think using the ternary to style is brilliant. Looks neat 👌

code/src/components/Feed.js
```diff
@@ -0,0 +1,86 @@
...
+
+import GeneratedFeed from 'components/GeneratedFeed.js';
+import PostNewThought from 'components/PostNewThought';
+import SiteHeader from 'components/SiteHeader';
+
+const Feed = () => {
+  const [thoughts, setThoughts] = useState([])
+  const [newThought, setNewThought] = useState('')
+  const [counter, setCounter] = useState(0)
+  const [loading, setLoading] = useState(false)
```
petrasson, line 12: It's hard to tell if the loading is working since the data loads fast, but I don't see anything that could be wrong. How do you know that it's not working?

Sneezan:
Thank you @petrasson for your feedback ☀️
You are right, the loading doesn't really have a function at the moment. I had issues with the loading making the page glitchy,  and at the time could not resolve that - which lead to the loading function getting omitted. I will revisit this later and see if it can be resolved, otherwise I'll remove it in it's entirety.  Thank you for getting back on my code and pointing out the console errors, I will check them out!

PR #350, opened 2022-10-07
Happy thoughts

CeciliaMichele:
Link to my deployed project:
https://thehappiestthougths.netlify.app/

This was very, very hard, the thoughest week so far I think. I struggled a lot with the like buttons and the structure of it all. 

I would have loved more help from the code coaches. The like buttons were really hard, and I would appreciate more presence from the teachers, since I felt I had to depend a lot on my classmates.


code/src/index.css
```diff
@@ -11,3 +11,78 @@ code {
...
+
+button:hover{
+  background-color: rgb(247, 132, 132);
+}
+ 
+
+.thought-text {
+  font-family: 'Inconsolata', monospace;
+  font-weight: 700;
+}
```
jessikalind, line 41: To make the long messages line-break, I added `overflow: auto` here. Like we talked about this week and on our Monday meeting, there are several ways of doing it though :)

jessikalind (commented):
Nice job on your Happy Thoughts, Cecilia! I know you had a tough week, but you have made a great application with a working API-connection and a super nice design (that is nearly identical to the design we were given for the assignment) 🥳

The main thing I looked at (and was trying to understand) was that your entire page reloads when liking a post. You can see my comment on this further down in the review and we can continue the discussion regarding this at a later point 😊

Great job!

PR #370, opened 2022-10-09
Niclas Jensen - Happy Thoughts

jnbjensen:
Link to my deployed project: 

https://jnbjensen-happy-thoughts.netlify.app/

Still catching up on React after taking a 2-week hiatus from Technigo, so I found this project very challenging. I'm happy with the end result, but I've relied a lot on peeking at other people's code. My focus was on ensuring that I properly understood how the app works, and I did so by adding thorough explanations for every block of code via comments.

I'm still struggling with APIs and React props in particular; I'll be spending more time on those two topics in the days to come.


hannanorgard (commented):
Hi Nick! As usual, super great job this week! Your commenting is always super clear and helpful, especially when I think I grasped something, I read your comments and they make me reflect a bit extra. Your website looks super clean, very nice design! Keep up the good work! :D

PR #372, opened 2022-10-09
Project-Happy Thoughts w7 Nina Berggren

NinaBerggren:
Link to my deployed project:
https://happy-thought-nina-berggren.netlify.app/


code/src/components/ListOfThoughts.js
```diff
@@ -0,0 +1,32 @@
...
+          <h4>{thought.message}</h4>
+          <div className="like-stats-container">
+            <button
+              type="button"
+              className={thought.hearts === 0 ? 'empty-btn' : 'like-btn'}
+              onClick={() => onThoughtLikeChange(thought._id)}>
+              <span className="heart" role="img" aria-label="heart symbol">
+              ❤️
+              </span>
+            </button>
```
LisaVasic, line 24: I like this piece of code. I was unsure how to make the like-button and make it change color. This together with the CSS is neat and easy to understand.

LisaVasic (commented):
Nice job with this project Nina!
The code is easy to understand and very neat with good comments.

Everything works and I would say it looks exactly like the given design.  I also like the subtle counter of characters which is a stretch goal. Very well done!

PR #386, opened 2022-10-09
Linnea happy thoughts

LinneaWa:
https://linnea-happy-thoughts.netlify.app/


code/src/index.css
```diff
@@ -1,13 +1,84 @@
...
+  padding: 15px;
+  line-height: 0%;
+  border-radius: 50%;
+  border: none;
+}
+
+.likeBtn:hover {
+  border: 1px solid gray;
+  cursor: pointer;
+}
```
MariaWellander, line 84: Also nice to add a hover effect here! Goes very well with the rest of the design here too :)

MariaWellander (commented):
**Am I able to understand the code easily? Does the code have lots of duplicates? Could it be tidied up?**

I'm absolutely able to understand the code easily - good job with that. It looks really neat and clean.

**Does it work or does anything look broken?**
	
It work's and nothing seems broken - well done!

**Does it follow the general and blue requirements?**

Yes, it does! Seems like some more requirements are also reached - awesome!

**To sum it up**

Wonderful work, Linnéa! You should be very proud of yourself :D High five!

PR #387, opened 2022-10-09
Project Happy Thoughts - Lisa J

LisaVasic:
Link to my deployed project:

### DIFF
--- a/code/src/components/ThoughtsList.js
+++ b/code/src/components/ThoughtsList.js
@@ -1,0 +1,35 @@
+import React from 'react';
+import moment from 'moment';
+/* import { formatRelative } from 'date-fns'; */
+
+// eslint-disable-next-line no-unused-vars
+const ThoughtsList = ({ loading, thoughtsList, setThoughtsList }) => {
+  if (loading) {
+    return <h1>Loading in progress...</h1>
+  }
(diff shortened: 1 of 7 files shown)


code/src/components/ThoughtsList.js
```diff
@@ -0,0 +1,32 @@
...
+// eslint-disable-next-line no-unused-vars
+const ThoughtsList = ({ loading, thoughtsList, setThoughtsList, onLikesIncrease }) => {
+  if (loading) {
+    return <h1>Loading in progress...</h1>
+  }
+
+  return (
+    <section className="thoughtWrapper">
+      <div>
+        {thoughtsList.reverse().map((thought) => (
```
lindanorberg, line 14: In the happy thought API (compared to the test API) the thoughts already comes in the right order. Therefore you don't need to reverse it.

lindanorberg (commented):
Great work Lisa! I know the individual weeks can be tough. 

I can´t find a deployed site but I know from the demo that you are almost done with this project. The website works so you load the latest thoughts and can push a new thought. The styling is almost done and you just have to add the function of pushing hearts to the API. 

The code is neat and tidy. I do understand the code and it is easy to read. 

Just reach out if you need help with somethings!

PR #395, opened 2022-10-15
Tara-project-happy-thought

Tara78:
Link to my deployed project:


jnbjensen (commented):
Great work Tara! I really like the comments that you added to make it easier to understand right away what each block of code does. My only suggestion would be to change some of the component names to match their function (i.e. TaskForm and TextList) to make it even clearer.  : )

PR #403, opened 2023-03-23
Leo T happy thoughts

khleoth:
Week 7 project including stretch goals: https://happythoughtsleot.netlify.app/

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,24 @@
-import React from 'react';
+/* eslint-disable no-underscore-dangle */
+import React, { useState, useEffect } from 'react';
+import { ThoughtsOutput } from 'components/ThoughsOutput';
+import { ThoughtsInput } from 'components/ThoughtsInput';
 
 export const App = () => {
+  const [thoughtList, setThoughtList] = useState([])
+  const [loading, setLoading] = useState(false)
+  useEffect(() => {
+    setLoading(true);
+    fetch('https://happy-thoughts-ux7hkzgmwa-uc.a.run.app/thoughts')
+      .then((response) => response.json())
+      .then((data) => setThoughtList(data))
+      .catch((error) => console.log(error))
+      .finally(() => { setLoading(false) })
+  }, []);
   return (
-    <div>
-      Find me in src/app.js!
+    <div className="main-container">
+      <ThoughtsInput />
+      {!loading && ThoughtsOutput(thoughtList)}
+      {loading && (<h2>LOADING</h2>)}
     </div>
   );
 }
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -6,17 +6,31 @@ import { ThoughtsInput } from 'components/ThoughtsInput';
 export const App = () => {
   const [thoughtList, setThoughtList] = useState([])
   const [loading, setLoading] = useState(false)
+  const submitHandler = (message) => {
+    if (message.length <= 5 && message.length >= 140) {
+      fetch('https://happy-thoughts-ux7hkzgmwa-uc.a.run.app/thoughts', {
+        method: 'POST',
+        body: JSON.stringify({ message }),
+        headers: { 'Content-Type': 'application/json' }
+      }).then((response) => response.json()).then((data) => {
+        setThoughtList((pv) => [data, ...pv])
+      })
+    } else {
+      alert('make sure the input is between 5-140 characters long')
+    }
+  }
+  /* Loads the 20 most recent happy thoughts from the API */
   useEffect(() => {
     setLoading(true);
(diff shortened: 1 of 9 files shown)


code/src/components/ThoughsOutput.js
```diff
@@ -0,0 +1,22 @@
...
+            <LikeButton id={thought._id} hearts={thought.hearts} />
+            <div className="date-display">
+              {formatDistance(new Date(thought.createdAt), Date.now(), { addSuffix: true })}
+            </div>
+          </div>
+        </div>
+      )
+    })
+  )
+}
```
jenster5, line 22: the console error suggests the date is undefined, maybe try to console log the error and see what you get from that since I cant really think why there would be an issue with the date value being passed into formatDistance when someone does not enter the right number of characters in the happy thoughts input.

code/src/App.js
```diff
@@ -1,9 +1,38 @@
...
+    if (message.length >= 5 || message.length <= 140) {
+      fetch('https://happy-thoughts-ux7hkzgmwa-uc.a.run.app/thoughts', {
+        method: 'POST',
+        body: JSON.stringify({ message }),
+        headers: { 'Content-Type': 'application/json' }
+      }).then((response) => response.json()).then((data) => {
+        setThoughtList((pv) => [data, ...pv])
+      })
+    } else {
+      alert('make sure the input is between 5-140 characters long')
```
jenster5, line 19: This alert is not working for me if I do less than 5 characters or more than 140 it goes to a blank page and has errors in the console log that mention App.js line 11 and ThoughtsOutput.js:line 8 and 15:16.

code/src/components/LikeButton.js
```diff
@@ -0,0 +1,34 @@
...
+        type="button"
+        className={`likeButton${likeButtonToggle}`}
+        disabled={like}
+        onClick={toggleLike}>
+        ❤️
+      </button>
+      <span>x {hearts}</span>
+    </div>
+  )
+}
```
jenster5, line 44: nice you figured out how to limit the clicks on the hearts i missed that in mine

code/src/components/SendButton.js
```diff
@@ -0,0 +1,12 @@
...
+export const SendButton = (message) => {
+  return (
+    <button
+      type="submit"
+      className="sendButton"
+      disabled={message.length < 6 || message.length > 140}>
+      ❤️Send a Happy Thought❤️
+    </button>
+  )
+}
```
jenster5, line 12: I wonder if this disabled is the source of your page breaking when you enter <5 or more than 140characters. what about something like this so its an if else statement.         
```
export const SendButton = ({ message }) => {
  const [isDisabled, setIsDisabled] = useState(true);

  const handleDisabled = (message) => {
    if (message.length < 6 || message.length > 140) {
      setIsDisabled(true);
    } else {
      setIsDisabled(false);
    }
  };

  handleDisabled(message);

  return (
    <button
      type="submit"
      className="sendButton"
      disabled={isDisabled}
    >
      ❤️Send a Happy Thought❤️
    </button>
  );
};
```

code/src/components/ThoughsOutput.js
jenster5: or see above I wrote this comment before I looked at your button

code/src/components/ThoughtsInput.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+  return (
+    <form onSubmit={onSubmit}>
+      <div className="inputBox">
+        <p>What&apos;s making you happy right now?</p>
+        <textarea
+          className="inputfield"
+          value={message}
+          onChange={(event) => setMessage(event.target.value)}
+          placeholder="React is making me happy!" />
```
jenster5, line 19: I like the place holder :)

jenster5 (commented):
Overall a really great job on the project it works to send a happy thought and to click on the hearts. I like how the hearts you click on are red so you can see which ones you have clicked on. It follows the design really well which is great. Well done!

code/src/components/ThoughsOutput.js
khleoth: Not sure what you mean here? I cant see anything wrong with the time stamp...? It works as intended and displays how long its been since its been published

jenster5: Sorry maybe i didnt explain well or I was to abstract. Yes the time stamp works well thats sort of what i was saying I am not sure why there is an error in the console referring to this line. So I think you can fix the problem with the page when someone types less than 5 characters or more than 140 and see what happens with the console output. Then maybe the error for the timestamp is gone.

code/src/App.js
khleoth: Thank you , did not notice this, sorted now!

PR #406, opened 2023-03-23
Andreas Happy Thoughts App

PerErikAndreas:


code/src/App.js
```diff
@@ -1,9 +1,117 @@
...
+  const [newMessage, setNewMessage] = useState('');
+
+  // Here we are calling the API and gets the JSON.
+  // The setLoading (true) shows "loading text" if the API call is delayed.
+  // .catch is catching eventual errors and displays them in a consol.log
+  // No options is created for this request its per default a GET
+
+  const fetchThoughts = () => {
+    setLoading(true)
+    fetch('https://happy-thoughts-ux7hkzgmwa-uc.a.run.app/thoughts')
```
malvinamaria: Instead of hard-coding the API endpoint https://happy-thoughts-ux7hkzgmwa-uc.a.run.app/thoughts in two places, define it as a constant and reuse it. This will make it easier to change the endpoint if needed.

code/src/App.js
```diff
@@ -1,9 +1,117 @@
...
+  const handleNewThoughtsChange = (event) => {
+    setNewMessage(event.target.value)
+  }
+
+  // This function takes an object as a parameter.
+  // The object in this case is the text the user writes
+  // The event.preventDefault prevents the default behavior for the form
+  // to refresh the page when pressing the submit button.
+
+  const onFormSubmit = (event) => {
```
malvinamaria, line 53: In the onFormSubmit function, you can use object destructuring to extract the newMessage variable instead of accessing it directly:` const { newMessage } = event.target.elements;`

code/src/Components/Footer/Footer.css
```diff
@@ -0,0 +1,64 @@
+.footer-box {
+    display: flex;
+    flex-direction: row;
+    align-items: flex-start;
+    width: 100%;
+    max-width: 270px;;
```
malvinamaria, line 6: In the CSS file Footer.css, there's a typo in the `max-width `property. Remove the extra semicolon after the value.

malvinamaria (commented):
Overall, the code looks good and is well-structured. It is easy to follow. Good job!

PR #407, opened 2023-03-23
EvaSjoholm/Happy-Thoughts-Project

EvaSjoholm:
The API is from Technigo and a shared one at this stage. 
Deployed page: 
https://spontaneous-beignet-bb56fb.netlify.app/


code/src/App.js
```diff
@@ -1,9 +1,81 @@
...
+      .then((res) => res.json())
+      .then((data) => setThoughts(data))
+      .finally(() => setLoading(false));
+  }
+
+  useEffect(() => {
+    fetchThoughts()
+  }, [])
+
+  const handleFormCleanup = () => {
```
khleoth: I recall this being mentioned but did not find use for it myslef. I take it that it is good practice but I fail to see it's function for this project...? Might just be that our code differ that much. :)

code/src/components/Spinner.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
```
khleoth, line 2: Nice feature, had to spam refresh page to see it myself :)

code/src/App.js
```diff
@@ -1,9 +1,81 @@
...
+      method: 'POST',
+      headers: {
+        'Content-type': 'application/json'
+      }
+
+    })
+      .then((res) => res.json())
+      .then(() => {
+        fetchThoughts()
+          .finally(() => handleFormCleanup())
```
khleoth: I can't really tell what's blocking the like-function but if I was to pick a guess I would say it's somewhere between line 56-58. Don't take my word for it though!

code/src/urls/urls.js
```diff
@@ -0,0 +1,3 @@
+// eslint-disable-next-line camelcase
```
khleoth, line 1: nice, this might be a good thing to take with me ahead, didn't think of it but it does tidy up a bit when handling multiple API's!

khleoth (commented):
Well done! I see that you have made an effort and started from scratch (twice!) yet still finished!

The code looks nice and tidy. I did not, however, get the like function to work and the heart itself is positioned differently than in the specs given, besides that it's just as intended.

Keep it up!

PR #414, opened 2023-03-24
Happy thoughts Emilia

emiliasaberski:


code/src/App.js
```diff
@@ -1,9 +1,79 @@
...
+  const [messageList, setMessageList] = useState([]);
+  const [loading, setLoading] = useState(false);
+  const [count, setCount] = useState(0);
+
+  useEffect(() => {
+    fetchMessages();
+  }, [newLike, loading]);
+
+  const fetchMessages = () => {
+    setLoading(false)
```
beemailley: Do you have somewhere where you set loading to "true"?  If not, do you need the loading/setLoading useState?

code/src/components/LikeBtn.js
```diff
@@ -0,0 +1,56 @@
...
+  const handleNewLikeChange = () => {
+    setNewLike(true)
+  }
+
+  // The function that registers the users like to the database, it also
+  // triggers the state-hook that makes the button change color
+  const onButtonClick = (like) => {
+    like.preventDefault();
+
+    if (liked === false) {
```
beemailley, line 19: I think having liked === false as a condition for being able to post likes is making it so that you can only like each thought once per page refresh.  

Could you do something with counting likes instead of using true/false?

code/src/components/newThoughts.css
```diff
@@ -0,0 +1,38 @@
+
```
beemailley, line 1: You had mentioned that you were struggling with formatting for responsiveness.  I wonder if the .thought-container and .new-thought-box classes are competing?  Also, maybe try a combination of width with view width (vw) as the unit of measurement and max-width to control for desktop sizing.

code/src/components/messageList.css
```diff
@@ -0,0 +1,53 @@
+
```
beemailley, line 1: Similar to the CSS for NewThought, I wonder if the .message-container and .message-box classes are competing? Also here, maybe try a combination of width with view width (vw) as the unit of measurement and max-width to control for desktop sizing.

beemailley (commented):
Overall, I think it looks good!  There were a few spots where I made comments either based on what you said you were still working on, or where I noticed a little bug.  But, all in all, good job!

PR #415, opened 2023-03-24
my project

AnnaElvine:
☀️


emiliasaberski:
I think everything looks good and works as expected. Since I barely understood what I was doing this week I'm sorry to say I don't feel like I can give you much relevant feedback. The app does what it should and I know how you struggled and think you should be really proud of your finished product! 🌟

PR #423, opened 2023-03-25
Jennifers-Happy-Page

jenster5:
https://jennifers-happy-thoughts-page.netlify.app/


code/src/App.js
```diff
@@ -1,9 +1,79 @@
...
+        message: newThought
+      })
+    }
+
+    fetch('https://happy-thoughts-ux7hkzgmwa-uc.a.run.app/thoughts', options)
+      .then((response) => response.json())
+      .then(() => fetchMessage())
+      .catch((error) => console.error(error))
+      .finally(() => setNewThought(''))
+  }
```
noraterbocs, line 45: First I thought the button didn't work but then I realized that it was only because 4 characters cannot be posted. You could add a conditional statement here in case the text is less then 5 characters then show an error message.

noraterbocs (commented):
I think you have done a good job, you kept the css and the components organized. If I could suggest you something is that you could give a hover effect and cursor:pointer for the interactive elements (button, heart emoji). Otherwise, well done! You've done a good job!

PR #429, opened 2023-03-26
https://dynamic-paprenjak-a7d590.netlify.app/

EmiliaGA:


code/src/App.js
```diff
@@ -1,9 +1,82 @@
-import React from 'react';
+/* eslint-disable max-len */
+/* eslint-disable spaced-comment */
```
YlvaKarlsson: I still think it's so funny that we should use ESLint and then a lot of us ends up disabling it to be able to deploy etc.. I did the same as you in some of my files :)

code/src/App.js
```diff
@@ -1,9 +1,82 @@
...
+      body: JSON.stringify({
+        message: newMessage
+      })
+    }
+
+    fetch('POST https://happy-thoughts-ux7hkzgmwa-uc.a.run.app/thoughts', options)
+      .then((res) => res.json())
+      .then(() => fetchThoughts())
+      .finally(() => setNewMessage(''));
+  }
```
YlvaKarlsson: I'm not sure if it's a bug here or if the Netlify is just weird tonight but when I try I can't post a new happy thought from the input field and make it show in the feed with happy thoughts.
I saw another way of solving it from a classmate and I'll post that solution here as it might help you figure out what's wrong (when I looked in the console after inspecting it also showed some sort of error with the POST request). I'm sorry I can't see or tell you exactly what's wrong as my own app is not even close to yours but I hope you can find something in the provided code-snippet below :) 

// Sending
  const postHappy = () => {
    fetch(APIEndpoint, {
      method: 'POST',
      body: JSON.stringify({ message: stateVariable }),
      headers: { 'Content-type': 'application/json; charset=UTF-8' }
    })
      .then(() => {
        setStateVariable(''); // Clear the input after the message has been sent to the API.
        fetchHappy(); // Update/Bring me the thought list with the new ones
      })
      .finally(() => {
        console.log('POST request was successful');
      })
  }

code/src/components/NewThought.js
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from 'react';
+
+export const NewThought = ({ newMessage, handleNewThoughtsChange, onFormSubmit }) => {
+  const [remainingChars, setRemainingChars] = useState(160 - newMessage.length);
+
+  const handleInputChange = (event) => {
+    const input = event.target.value;
+    const remaining = 160 - input.length;
```
YlvaKarlsson: Nice styling to the remaining characters - very sophisticated and smooth with the faded grey color - Good job!

After writing in the input field - the text disappears but the character counter remains. So if I wrote 10 letters - they disappear (without showing the new happy thought) and the counter remains the same.
Only a hard update of the page sets the character counter back to 160, my new happy thought is still not showing up though.

I think it is some small detail in these lines with the handleInputChange that needs something - I'm sorry I can't tell you exactly how to fix it.
I saw another student who put the character-counter in their code like this and the name is different from your code but their "newThought" is the one you call "newMessage" so that souldn't matter.
This student had the fetch for the New Thoughts in this component so it might be an idea to refactor your code this way?
Just an idea - as I mentioned several times I'm not sure exactly whats wrong so I can just give some hints and tips from others code where I can see it works.
I think you are on the right track and that it's just minor details that will solve the bugs in the code.
Here is the example code from another student regarding the character-counter (when the fetch is starting further down I didn't copy-paste the whole fetch since that one is structured like yours but the things above it might be an idea to try out for your New thought-feed? 🥇 

 const HandleFormSubmit = (event) => {
    event.preventDefault();
    if (newThought.length < 5) {
      return alert('Get on that keyboard, more than 5 characters, please!')
    } else if (newThought.length > 140) {
      return alert('Are you some sort of hacker?! No more than 140 characters, please!')
    } else {
    const Submit = {
        method: 'POST',
        headers: { 'Content-Type': 'application/json' },
        body: JSON.stringify({ message: newThought })
      };
      fetch('https://happy-thoughts-ux7hkzgmwa-uc.a.run.app/thoughts', Submit)

code/src/components/ThoughtList.js
```diff
@@ -0,0 +1,32 @@
...
+
+import React from 'react';
+import { formatDistance } from 'date-fns';
+
+export const ThoughtList = ({ loading, thoughts, onLikesIncrease }) => {
+  if (loading) {
+    return <div className="Loading">Loading</div>
+  }
+
+  const handleOnLikesIncrease = (id) => {
```
YlvaKarlsson: **Very good job with the like-button. It works like a charm! =)** 
What would add something extra to your app might be to change the color on the button after a happy thought has gotten a like - but that's just a nifty extra feature and no requirements in this weeks project :) 
I saw that you button works with the hover-effect if it has no likes already but it might be an idea to keep unliked thoughts with a grey background on the button and change the ones that has likes to something else - just an idea on how to improve your app even more.

YlvaKarlsson (approved):
Overall very clean and neat code - both in the JS-files and the CSS - easy to follow and great work with commenting a lot.

There are some minor bugs that affects the functionality but I think you can solve these with ease and hopefully some of my tips can guide you in the right direction.
I think you should be very proud after that last week - it was though for many and VERY tough for me. I think you're on a good path to solving this project and I'm so happy to follow your imporvement in the team.
I'm sorry I can't pinpoint the exact issues that causes the bugs in your code but at least I hope this CR gives some ideas for what to look up to solve the things that aren't working at the moment.

Great work week 7 and let's rock on in week 8.
See you! :)

PR #440, opened 2023-03-26
Nora Terbocs - Happy thoughts

noraterbocs:
noras-happy-thoughts-project.netlify.app


code/src/App.js
```diff
@@ -1,9 +1,97 @@
...
+    const options = {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        message: newThought
+      })
+    }
+    if (newThought.length > 4) {
```
Joanna-Philips, line 39: It seems like this if condition is not fully functional, because a  new thought is still posted if the length = 4.

code/src/App.js
```diff
@@ -1,9 +1,97 @@
...
+      .finally(() => { setLoading(false) })
+  }
+  useEffect(() => {
+    fetchThoughts()
+  }, [])
+  console.log(errorMsg)
+  const handleFormSubmit = (event) => {
+    console.log(newThought)
+    event.preventDefault()
+    const options = {
```
Joanna-Philips, line 30: Good idea storing the method in a variable.

code/src/components/Loading.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import FadeLoader from 'react-spinners/FadeLoader'
+
+const Loading = () => {
+  return (
+    <section>
+      <FadeLoader />
```
Joanna-Philips, line 7: Nice addition of the loader

Joanna-Philips (changes requested):
Excellent job on this week's project. :D

What went well:
- The code is well organised and easy to read. 
-Stretch goals done well. 
-The page is responsive, and functional. 

What could be better:
-The if condition for the length of the new thought doesn't work fully - see comment in code.

As always, really nice work Nora. Keep it up! 🎉

code/src/App.js
noraterbocs: I think that shouldn't be possible because even without adding an if statement the post request blocks it if the text is under 5 characters. The if statement only adds the error msg.

PR #449, opened 2023-03-26
Project happy thoughts

SofiaGerdmar:


code/src/App.js
```diff
@@ -1,9 +1,52 @@
...
+  const [loading, setLoading] = useState(false)
+
+  const fetchThoughts = () => {
+    setLoading(true)
+    fetch('https://happy-thoughts-ux7hkzgmwa-uc.a.run.app/thoughts')
+      .then((response) => response.json())
+      .then((data) => setThoughtsList(data))
+      .catch((error) => console.log(error))
+      .finally(() => { setLoading(false) })
+  }
```
JohLeo: To keep the App component cleaner (not really necessary I think in this size of projects but a good practise for the future when we work in bigger projects) this could've been places in one of your components. Since this is fetching the thoughts from the API it would be neat to place it in your ListThoughts.js maybe!

code/src/App.js
```diff
@@ -1,9 +1,52 @@
...
+
+  useEffect(() => {
+    fetchThoughts()
+  }, [])
+
```
JohLeo, line 24: This hook is to call the fetchThoughts function when the component is mounted so this should/could/would also be moved if so we were to change the structure for a "cleaner" App.js

code/src/App.js
```diff
@@ -1,9 +1,52 @@
...
+  const handleFormSubmit = (event) => {
+    event.preventDefault()
+    setLoading(true)
+
```
JohLeo, line 32: Since the "handleFormSubmit" handles the submission of a new thought it could be moved into your component "submitThoughts.js" instead if we want to continue on "cleaning" the app.js a bit.

code/src/components/ListThoughts.js
```diff
@@ -1,13 +1,36 @@
...
+              </button>
+              <p>x {thought.hearts}</p>
+              <div>
+                <p className="time">{getTimeAgo(thought.createdAt)}</p>
+              </div>
+            </div>
+          </div>
+        )
+      })}
+    </section>
```
JohLeo, line 34: Very nice and tidy indentation, pretty code!

code/src/components/SubmitThoughts.js
```diff
@@ -1,10 +1,34 @@
...
-  </form>
+export const SubmitThoughts = ({ handleFormSubmit, submitThoughts, handleThoughtChange }) => {
+  const turnCounterRed = submitThoughts.length > 140 ? 'counter-red' : 'counter';
+  const [errorAlert, setErrorAlert] = useState('')
+  const handleLengthError = (event) => {
+    const textLength = event.target.value.length
+    if (textLength < 5) {
+      setErrorAlert('Your message needs to be at least 5 characters long. 😊')
+    } else if (textLength > 140) {
+      setErrorAlert('Whoa Nelly! Save the essay for your portfolio. 😄')
```
JohLeo, line 12: I wish I tried writing a message that was to long, fuunny! 😂

code/src/index.css
```diff
@@ -1,13 +1,192 @@
...
+
+.counter-red {
+  color: red;
+  text-align: right;
+  font-size: 12px;
+}
+
+.errorAlert {
+  font-size: 12px;
+  margin-bottom: 10px;
```
JohLeo, line 80: Oh so the alert pops up in the thoughstlist instead of beeing an actual alert! Is that on purpose?

code/src/index.css
```diff
@@ -1,13 +1,192 @@
...
+  box-shadow: none;
+  margin-right: 5px;
+  margin-bottom: 10px;
+}
+
+.thought {
+  margin-left: 10px;
+  font-family: 'Courier New', Courier, monospace;
+  font-weight: 500;
+}
```
JohLeo: How and where in you code are you making the text of this card to not overflow if it where an textsnippet that was to long? In the inspect there is a "overflow-wrap: anywhere;"  but i cannot find it and how you did it. Is it magic? Curios to know!

JohLeo (approved):
Well done!! Your project is so pretty and I loove the colours!  One tiny improvement would be to split the logic for fetching and submitting thoughts into their own functions outside of the App component. That would make the code more modular and easier to test. 
The App component can then call these functions as needed. The(very nice) loading spinner can be placed in its own componen as well, which can be reused throughout the application!
But that is just a preference!    Be proud you have done a great job, again!! 🤩☀️

PR #451, opened 2023-03-26
TOTAL DISASTER - NOT working at all - W/ - Sad Thoughts and not a single happy one /Ylva K

YlvaKarlsson:
So sad to say that this week is a total failure for me.


EmiliaGA (commented):
Good job staying sane this week! It's been a tough one for all of us and your project is not at all a complete failure... You've managed to fetch the API and it's looking pretty close to the original, that's something to be happy about I think :) As far as I can tell it's the posting-function not working quite properly. The like-button seems to work, perhaps add some hover-effects/change of background color when liked/not liked. Otherwise, the code looks nice and you're close to finishing :D Be proud!

